PR #6, opened 2018-04-18
 Add a --keep-v option to not drop `v` at the beginning of version numbers

samoht:
/cc @diml 

I've done a minimal amount of testing with sexplib0 with `dune-release distrib --keep-v` and `dune-release opam --keep-v` and that seems to work fine.

### DIFF
--- a/bin/bistro.ml
+++ b/bin/bistro.ml
@@ -6,14 +6,15 @@
 
 open Bos_setup
 
-let bistro () =
+let bistro () keep_v =
   begin
     let verb = Cmd.(v "--verbosity" % Logs.(level_to_string (level ()))) in
+    let args = if keep_v then Cmd.(verb % "--keep-v") else verb in
     let dune_release = Cmd.(v "dune-release") in
-    OS.Cmd.run Cmd.(dune_release % "distrib" %% verb)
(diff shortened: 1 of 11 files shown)


diml:
Thanks!

One thing I'm thinking is that keeping the v prefix is a property of the package. Would it be possible for this setting to be persistent? For instance by writing `x-keep-v-prefix: true` in th opam file

diml:
Another idea: we are currently moving forward with the renaming (https://github.com/ocaml/dune/pull/721) and have started adding support for `dune-project` files. There will be a `(version ...)` stanza in this file. Whether we want the `v` prefix or not could be derived from this stanza. For instance you'd write `(version %%VERSION%%)` (the default) to keep the `v` prefix and `(version %%VERSION_NUM%%)` to drop it.

We can expose and API in dune to parse this file so that dune-release can access this information easily. What do you think?

samoht:
> We can expose and API in dune to parse this file so that dune-release can access this information easily. What do you think?

That seems a good idea. Having `x-keep-v-prefix: true` looks like a hack and I am not sure opam 1.2 will be happy with that.

samoht:
Merging from now, we will look at how to persist the option later.

Issue #4, opened 2018-04-13
Issue when releasing from a LTS branch


samoht:
If the tag we are using while doing `dune-release publish submit` is not at the tip of master, `toy-github-delegate` (or its equivalent, eg. `Dune_release.Github.publish_distrib`) becomes confused a creates a new release on the wrong commit..

PR #24, opened 2018-06-07
publish: do not force push all the tags to a random repo

samoht:
Instead, just force-push the latest tag to the dev-repo. Force-pushing
is probably not what we want here, but that's already an improvement.


samoht:
This should fix #4

PR #32, opened 2018-06-13
A few fixes as I use dune-release for the first time

avsm:


samoht:
Can you revert e4bd7ad ? I think we should stick to dune conventions here.

avsm:
It's necessary to make the tool work at all -- without it, I get an Invalid_argument raised from cmdliner due to duplicate Arg registrations. Merge this and then remove the Webbrowser_cli clash would work I think.

PR #41, opened 2018-07-08
A few bug-fixes/improvements

let-def:
The implementation is crude, you might want to revisit the code.
Changes:
1) bug fix, `drop_initial_v` in `opam publish` too (the third copy of this code after distrib.ml and pkg.ml)
2) ux improvement, basic token validation
3) ux improvement(?), echo token on stdin when copying it. I don't think this is critical for security, I am curious what your opinion is :)


samoht:
Looks good, thanks!

PR #46, opened 2018-07-09
Improve heuristic to guess the package main name

### LINKED ISSUE #33: multi-name opam files are problematic
The automatic name detection fails to work for repos like https://github.com/ocaml/obi that do not have a common-short-prefix.

For these, having some convention for metadata (descr) in the repo would be handy, so that we don't need to specify a CLI flag all the time.

samoht:
Fixes #33

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -300,9 +300,28 @@ let infer_name () =
     if List.for_all (String.is_prefix ~affix:shortest) package_names
     then shortest
     else begin
-      Logs.err (fun m ->
-          m "cannot determine name automatically. Use `-p <name>`");
-      exit 1
+      let err () =
+        Logs.err (fun m ->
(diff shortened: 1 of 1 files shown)


rgrinberg:
Note that the package name is now available in the `dune-project` file. dune-release shouldn't rely on any heuristics anymore.

samoht:
See https://github.com/samoht/dune-release/blob/master/lib/pkg.ml#L333-L344

Is there an easier way to get that package name?

rgrinberg:
Unfortunately, there is not. This is one of those things that would be quite a bit easier if dune-release was part of dune. In any case, we can solve this problem without considering any kind of merger between the two, as I'm sure other tools are going to want to solve this problem. Here's a quick proposal on how to address this issue https://github.com/ocaml/dune/issues/1279

PR #52, opened 2018-07-19
Fix typo in error message

rgrinberg:
-p no longer exists. The option was renamed to -n

Signed-off-by: Rudi Grinberg <rudi.grinberg@gmail.com>


samoht:
I've fixed the option name instead :-)

PR #62, opened 2018-08-14
Remove short option p of opam sub command line (conflict)

dinosaure:
Got an `Invalid_argument` when `-p` is defined two times on the `opam` sub-command line.

### DIFF
--- a/bin/opam.ml
+++ b/bin/opam.ml
@@ -290,7 +290,7 @@ let pkg_names =
              by the package description."
   in
   let docv = "PKG_NAME" in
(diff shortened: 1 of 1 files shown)


samoht:
Do you known what is the the other option it is conflicting with?

dinosaure:
We have a conflict with `Cli.pkg_name` and `Opam.pkg_name` on `Opam.cmd` (`bin/opam.ml`, l. 323 and l. 325)

samoht:
See #67

Issue #60, opened 2018-08-08
Final pull request to opam-repository fails


mor1:
Not sure what's happening here. Seem to get a failure when `dune-release opam submit --pkg-names ...` submits a multi-package (single repo) PR to opam-repository. I appear to be able to create the PR myself by hand, no problem.

Error is

```
...
Total 17 (delta 6), reused 0 (delta 0)
remote: Resolving deltas: 100% (6/6), completed with 3 local objects.
To github.com:mor1/opam-repository
 + 666d75518f...230beffc54 release-owl-0.4.0 -> release-owl-0.4.0 (forced update)
dune-release: [ERROR] Could not find html_url id in response:
                      HTTP/1.1 422 Unprocessable Entity
                      Server: GitHub.com
                      Date: Wed, 08 Aug 2018 23:15:19 GMT
                      Content-Type: application/json;
                      charset=utf-8
                      Content-Length: 235
                      Status: 422 Unprocessable Entity
                      X-RateLimit-Limit: 5000
                      X-RateLimit-Remaining: 4830
                      X-RateLimit-Reset: 1533771756
                      X-OAuth-Scopes: public_repo
                      X-Accepted-OAuth-Scopes:
                      X-GitHub-Media-Type: github.v3;
                      format=json
                      Access-Control-Expose-Headers: ETag, Link, Retry-After,
                      X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining,
                      X-RateLimit-Reset, X-OAuth-Scopes,
                      X-Accepted-OAuth-Scopes,
                      X-Poll-Interval
                      Access-Control-Allow-Origin:
                      *
                      Strict-Transport-Security: max-age=31536000;
                      includeSubdomains; preload
                      X-Frame-Options: deny
                      X-Content-Type-Options: nosniff
                      X-XSS-Protection: 1; mode=block
                      Referrer-Policy: origin-when-cross-origin,
                      strict-origin-when-cross-origin
                      Content-Security-Policy: default-src
                      'none'
                      X-Runtime-rack: 0.073096
                      X-GitHub-Request-Id:
                      D709:09BB:2407C4:534A9D:5B6B7986

                      {
                        "message": "Validation Failed",
                        "errors": [
                          {
                            "resource": "PullRequest",
                            "field": "head",
                            "code": "invalid"
                          }
                        ],
                        "documentation_url":
                      "https://developer.github.com/v3/pulls/#create-a-pull-request"
                      }
                      .
```

[...]

PR #64, opened 2018-08-29
 Determine opam-repository fork user from URI

NathanReb:
# What's wrong

When running `dune-release opam submit`, the actual PR creation on `opam-repository` can fail due to an invalid `head` field in the API requests. As far as I understand, this appears to happen when the user or organisation owning the package repo differs from the `opam-repository` fork owner. See #60 for further details.

# The actual fix

Since `dune-release opam submit` knows the `opam-repository` fork URI, it can be used to determine the owner of the fork and pass it in the github API request instead of the package repository owner.

# Notes

This is a first draft, I mostly wanted to have your feedback on the fix.

I wasn't sure whether HTTP URIs should be supported along with SSH ones here but that should be fairly easy to add.

If the parsing of the URI fails it falls back to `user`.

You'll also note I added some tests in a separate commit.

Let me know what you think!

Thanks

### DIFF
--- a/bin/opam.ml
+++ b/bin/opam.ml
@@ -100,8 +100,8 @@ let submit ~dry_run opam_pkg_dir local_repo remote_repo pkgs =
         strf "[new release] %a (%s)" (pp_list Fmt.string) names version
       in
       Pkg.publish_msg pkg >>= fun changes ->
-      Pkg.distrib_user_and_repo pkg >>= fun (user, repo) ->
-      let changes = rewrite_github_refs user repo changes in
+      Pkg.distrib_user_and_repo pkg >>= fun (distrib_user, repo) ->
+      let changes = rewrite_github_refs distrib_user repo changes in
       let msg = strf "%s\n\n%s\n" title changes in
       Opam.prepare ~dry_run ~msg ~local_repo ~remote_repo ~version names
       >>= fun branch ->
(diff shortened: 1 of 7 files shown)


NathanReb:
Fixed the CI in #65 !

samoht:
That looks great, thanks! 

A few remarks:
- the `opam_fork_user` variable is a bit confusing: I had to look at the code again to understand how it's different from the already existing`user`; so maybe the right fix would be to rename `user` to something better (maybe `distrib_user`?) to be able to use `user` instead of `opam_fork_user`. What do you think?
-  The additions of the tests are very welcome, but I would prefer if they use alcotest instead of ounit, if possible. I'll be very happy to add more tests later :-)

NathanReb:
I renamed the `user` and `opam_fork_user` variables and parameters and switched to `alcotest`!

I'm an `alcotest` virgin and because it lacks ounit nesting I'm torn between having a `test_case` per `check` and the current implem. I personally feel like it should be the former but without the nesting it doesn't feel so great and it makes the test executable output very verbose so I went for the latter!

samoht:
Looks great now, thanks!

samoht:
> Also realized dune-release isn't built with dune yet, happy to contribute with the upgrade as well!

That would be great!

PR #88, opened 2018-09-23
Updates to use opam2 release workflow

samoht:


lib/github.ml
```diff
@@ -231,7 +231,7 @@ let curl_open_pr ~token ~dry_run ~title ~distrib_user ~user ~branch ~body curl =
   let repo = "opam-repository" in
   let uri = strf "https://api.github.com/repos/%s/%s/pulls" base repo in
   let data =
-    strf {|{"title": %S,"base": "master", "body": %S, "head": "%s:%s"}|}
+    strf {|{"title": %S,"base": "2.0.0", "body": %S, "head": "%s:%s"}|}
```
anuragsoni: i thought that the master branch now defaults to the `2.0.0` format?

samoht: I thought the `master` branch was supposed to disappear -- I'll check the status with other maintainers.

anuragsoni: I was referring to the blog post for the official 2.0 release: https://opam.ocaml.org/blog/opam-2-0-0/

samoht: so you were right, PRs should continue to be opened against master. I've reverted that commit.

PR #95, opened 2018-09-25
Remove the status and log subcommands

samoht:
This is a tentative to only keep the main commands in `dune-release` to make a release. /cc @rgrinberg who suggested this.


rgrinberg:
Looks good. Thanks

Issue #94, opened 2018-09-25
doc publishing doesn't copy files to the root level of gh-pages


anuragsoni:
Maybe i missed some configuration, but during publishing `dune-release` is copying the `_html` folder into the root of gh-pages branch. Since there is no `index.html` in the directory github pages doesn't pick it up.

Example: https://github.com/anuragsoni/color/tree/95ab98acfd4ba3fe7862d4b3f9d9d7b7d3f40824

samoht:
Which version of `dune-release` are you running? I thought I fixed this but apparently not...

It still not very clear to me when `odoc` is creating a subdir and when it isn't. /cc @Anton and/or @rizo for help :-)

anuragsoni:
That particular commit was created with dune-release 1.0

I tried with odoc `1.2.0` and by pinning it to master but its the same behavior in either case. I can try taking a look at this if someone can point where to start 😄. I don't know too much about how odoc works.

samoht:
The function doing the build and copying the result to the right place is https://github.com/samoht/dune-release/blob/master/bin/publish.ml#L10-L20

aantron:
@samoht The directory structure is created by Dune, not odoc, so cc @rgrinberg.

The directory structure has changed at some point in the development of Dune/Jbuilder, but I am not very familiar with the details. Not sure if this is the problem here.

rgrinberg:
Yes but the change was ages ago and the old structure lasted for maybe 1 point release of dune. In any case, this is yet another argument for making dune-release a part of dune. One less thing to keep in sync :)

anuragsoni:
https://github.com/anuragsoni/color/commit/f209a500112fbf4ae267bcfdb67d2947aa2e053f

Adding a folder name in the doc url seems to help. Github pages can pick it up since the foldername doesn't start with an underscore.

anuragsoni:
The value for the documentation URI needs to be of the form "$SCHEME://$USER.github.io/$REPO/$PATH". When setting the PATH to an empty value, `dune-release` is just copying the `_html` folder without renaming it. Github-pages doesn't pick it up for reasons I described above. Should a default value for `PATH` be picked so we don't up in such a situation?

This could also be a candidate for a check that can be added to the `lint` step as picking a path that starts with an underscore will not work with github pages or any other static site host that uses jekyll.

samoht:
I think this should be fixed by #97 as well

anuragsoni:
[...]

PR #100, opened 2018-09-29
[WIP] Add lint for missing doc folder

anuragsoni:
Relates to #94 

Instead of a default, i was thinking a lint might be nicer after all. I think keeping the URL from the `opam` file is the best idea.

I'm not sure if the lint message i've added is the best option. I'm happy to reword.
As for tests, what do you think about adding `craml` or something similar to test the CLI ?

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -311,6 +311,11 @@ let doc_user_repo_and_path p =
             field value %a; expected the pattern \
             $SCHEME://$USER.github.io/$REPO/$PATH" String.dump uri
   in
+  (* If path is left empty, the docs are copied to a directory
+     called _html, which won't be picked up by github pages. *)
+  let path_error =
+    R.msgf "Publication directory $PATH should not be an empty string."
+  in
   match Text.split_uri ~rel:true uri with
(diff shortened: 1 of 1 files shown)


anuragsoni:
Not sure if adding `craml` would be a little tedious here since `dune-release` expects to work on git repositories.

anuragsoni:
`dune-release` runs `cp -R   /home/<some-dir>/_build/<distrib-folder>/_build/default/_doc/_html/ .` when its copying the documentation files and the path is `.` Atleast on linux it just copies the folder `_html` to the root.

anuragsoni:
I'll try to get to this today/tomorrow. I am not sure removing the trailing `/` or using the absolute path will fix this as the folder name will still remain `_html`. I think the behavior we want is `cp -R _html/* .` right? I'm not sure if `Fpath` will let me construct a path with wildcards like that.

anuragsoni:
I'm guessing this can be closed now that #94 is merged?

samoht:
Indeed. Thanks anyway for your detective work on the issue, that was very useful!

PR #129, opened 2019-02-20
embed a 'version: "$release-version"' in each opam file of the current directory when building an archive

### LINKED ISSUE #128: embed version in all opam files
At the moment, only the "main" opam file (e.g. `mirage.opam` for the mirage repositories) gets a `version: "YYY"` embedded during `dune-release distrib`. `dune` uses the version information from the opam file to embed it into the `META` file etc..  Other opam packages except the main one in the same repository (e.g. `mirage-types`) do not embed version information, and thus `$ ocamlfind query -format %v mirage-types` returns `[unspecified]`.

This issue was reported as https://github.com/ocaml/dune/issues/1844, but it seems to be a dune-release issue, thus reporting here instead.

hannesm:
fixes #128

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -445,7 +445,7 @@ let v ~dry_run
 
 (* Distrib *)
 
-let distrib_version_opam_files ~dry_run _p ~version =
+let distrib_version_opam_files ~dry_run ~version =
   infer_pkg_names Fpath.(v ".") [] >>= fun names ->
   List.fold_left (fun acc name ->
       acc >>= fun _acc ->
@@ -464,7 +464,7 @@ let distrib_prepare ~dry_run p ~dist_build_dir ~name ~tag ~version ~opam =
   Sos.with_dir ~dry_run dist_build_dir (fun () ->
       Distrib.files_to_watermark d ()
       >>= fun files -> Distrib.watermark_files ws_defs files
-      >>= fun () -> distrib_version_opam_files ~dry_run p ~version
+      >>= fun () -> distrib_version_opam_files ~dry_run ~version
       >>= fun () -> Distrib.massage d ()
       >>= fun () -> Distrib.exclude_paths d ()
     ) ()
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -445,13 +445,17 @@ let v ~dry_run
 
 (* Distrib *)
 
-let distrib_version_opam_files ~dry_run p ~version =
-  opam p
-  >>= fun file -> OS.File.read_lines file
-  >>= fun o ->
-  let o = List.filter (fun l -> not (String.is_prefix ~affix:"version" l)) o in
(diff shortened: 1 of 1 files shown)


hannesm:
when using `dune-release distrib` in the mirage/mirage git repository, and comparing the produced tarballs, I see that the version is embedded in all opam files, as intended, after this PR -- see https://try.diffoscope.org/abepfxtzqpgq.html

lib/pkg.ml
```diff
@@ -445,13 +445,17 @@ let v ~dry_run
...
 (* Distrib *)
 
-let distrib_version_opam_files ~dry_run p ~version =
-  opam p
-  >>= fun file -> OS.File.read_lines file
-  >>= fun o ->
-  let o = List.filter (fun l -> not (String.is_prefix ~affix:"version" l)) o in
-  let o =  Fmt.strf "version: \"%s\"" version :: o in
-  Sos.write_file ~dry_run file (String.concat ~sep:"\n" o)
+let distrib_version_opam_files ~dry_run _p ~version =
```
samoht: not sure about this -- the package list can be overriden by the `--packages` option so it's best to call `infer_pkg_names` only in one place in the binary (if `--package` is not set). Could you instead change `p` to be a list of packages and propagate the changes in the rest of the code?

samoht:
btw, the diffoscope link is broken.

hannesm:
Hello, sorry my late reply - I somehow missed that you commented here.

> btw, the diffoscope link is broken.

from their [website](https://try.diffoscope.org): "Results are purged after 30 days". I re-uploaded similar tarballs again at https://try.diffoscope.org/sypxaudhedyb.html

> not sure about this -- the package list can be overriden by the --packages option so it's best to call infer_pkg_names only in one place in the binary (if --package is not set). Could you instead change p to be a list of packages and propagate the changes in the rest of the code?

I'm not so sure about this. Current `master` behaves that it modifies only the "main" package opam file by embedding a `version`. This PR goes along and embeds into _all_ opam files a `version`. These opam files are never committed anywhere, but are only part of the tarball.

The resulting tarball has branding and watermarking (%%VERSION%% etc.) applied -- independent of which packages (using `-p`) are specified. Or, the other way around: should `dune-release` produce different tarballs for given `-p` arguments? In practice, I prefer to have _reproducible tarballs_ - independent of `-p` command-line arguments. This allows a `dune-release -p p1` followed by a `dune-release -p p2`, using the _identical_ tarball, and preserving `ocamlfind` version information.

samoht:
> This allows a dune-release -p p1 followed by a dune-release -p p2, using the identical tarball, and preserving ocamlfind version information.

This makes perfect sense! So instead could you remove the `_p` argument which is not useful anymore? Thanks!

hannesm:
@samoht 
> could you remove the `_p` argument

done in 059f20c

samoht:
Thanks!

PR #131, opened 2019-03-09
generate sha256 and sha512 checksums for release

hannesm:
use opam-core for doing these operations. instead of generating an URL and
checksum string (and parse with OpamFile.URL.read_from_string), construct a
OpamFile.URL.t directly in Opam.URL.with_distrib_file.

since dune-release always generates opam2 files, I don't see any reason to embed the md5 checksum in there at all...


samoht:
Looks good, thanks!

emillon:
A bit late but a bunch of crypto-thanks for this!

PR #133, opened 2019-03-29
handle doc fields with no trailing slash

yomimono:
`opam lint` doesn't mind doc fields without trailing slashes, so it seems to me that `dune-release` shouldn't either.


samoht:
Looks good, thanks!

PR #135, opened 2019-05-13
Add logs to describe the tag command behaviour

NathanReb:
This PR adds some extra logs to the `tag` command to describe the tag selection steps as it goes.

This is the first step in an effort to try and make `dune-release` output a bit more descriptive for the end users. I decided to tackle the simplest command first, I'll open separate PRs for the others.

Let me know what you think!


samoht:
Thanks! Something else which might be useful would be to ask user before doing any modification to the local repository (like creating a new tag)

PR #140, opened 2019-05-15
Allow to submit packages to a different opam repo on github

NathanReb:
I'm not sure this should be merged but it is really convenient for testing. I guess it could be useful for people maintaining custom opam repositories as well.

Let me know what you think!


samoht:
Good idea!

PR #137, opened 2019-05-13
Add logs to describe the publish command behaviour

NathanReb:
This PR adds some app level logs to describe what's going on during a `publish` run.

Let me know what you think!

PR #141, opened 2019-05-15
Add logs to describe the opam command's behaviour

NathanReb:
Depends on #139 

This PR adds a few extra logs to describe the detailed steps of `opam pkg` and `opam submit` sub-commands.
It also silence some of the command being run by using `Sos.run_quiet` in a similar fashion to wha was done in #137.

This is what the output looks like (tested on top of #140):
Screenshot from 2019-05-15 18-54-49
Screenshot from 2019-05-15 18-37-06

### DIFF
--- a/bin/opam.ml
+++ b/bin/opam.ml
@@ -24,7 +24,7 @@ module D = struct
   let distrib_uri = "${distrib_uri}"
 end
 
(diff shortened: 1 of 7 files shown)


samoht:
Looks great, thanks!

PR #143, opened 2019-05-21
Use the archive download URL returned by Github's API

### LINKED ISSUE #142: Wrong distrib archive name if release asset is renamed by Github
See https://github.com/ocaml/opam-repository/pull/14156

Based on the documentation [here](https://developer.github.com/v3/repos/releases/#upload-a-release-asset) the issue is that the github release API will rename asset that contain special characters (in this case the `+`).

I would be happy to enhance `dune-release` to query the asset names once it has uploaded them, but could do with some guidance on the best way to handle this - should it be done as part of asset upload and stored somewhere or should it be part of the opam release command?

NathanReb:
This fixes #142. We use the returned browser_download_url instead of generating it ourselves. This fixes the issues with sanitized URLs and filenames witout having to maintain sanitization code in `dune-release`.

The URL is stored in a `_build/<distrib_archive_base_filename>.url` file when running `dune-release publish` during the distribution pubishing.

Later on, during `dune-release opam pkg`, if that file exists we use the URL it contains otherwise we fallback to the old way. That makes it easier to keep working with custom delegates for now, although I'm not sure we really want to support them for long.


NathanReb:
You'll note the fragile custom json response parsing but I didn't want to introduce a JSON parser dependency, especially if we're considering merging the tool to `dune`.

NathanReb:
cc @leamingrad

leamingrad:
@NathanReb: This seems to have worked correctly for zxcvbn - thanks for the fix!

PR #146, opened 2019-05-27
Add more prompts

NathanReb:
This PR cleans up a few things, improve the logs in some places and add prompts before critical actions.

### DIFF
--- a/bin/opam.ml
+++ b/bin/opam.ml
@@ -116,7 +116,8 @@ let pp_opam_repo fmt opam_repo =
   Format.fprintf fmt "%s/%s" user repo
 
 let open_pr
-    ~dry_run ~changes ~remote_repo ~distrib_user ~branch ~token ~title ~opam_repo ~auto_open pkg =
+    ~dry_run ~changes ~remote_repo ~distrib_user ~branch ~token ~title ~opam_repo ~auto_open ~yes
+    pkg =
   Pkg.opam_descr pkg >>= fun (syn, _) ->
   Pkg.opam_homepage pkg >>= fun homepage ->
   Pkg.opam_doc pkg >>= fun doc ->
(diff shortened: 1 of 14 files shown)


bin/opam.ml
```diff
@@ -191,6 +170,34 @@ let submit ~dry_run ~opam_repo local_repo remote_repo pkgs auto_open =
...
+let submit ~dry_run ~yes ~opam_repo local_repo remote_repo pkgs auto_open =
+  Config.token ~dry_run () >>= fun token ->
+  List.fold_left (fun acc pkg ->
+      get_pkg_dir pkg
+      >>= fun pkg_dir -> Sos.dir_exists ~dry_run pkg_dir
+      >>= function
+      | true  -> acc
+      | false ->
+          Logs.err (fun m ->
+              m "Package@ %a@ does@ not@ exist. Did@ you@ forget@ \
```
avsm: I think Fmt takes care of the `@ ` space mapping for you automatically as part of Logs, so you can just use a normal space

lib/github.ml
```diff
@@ -293,17 +298,25 @@ let publish_distrib ~dry_run ~msg ~archive p =
   >>= fun tag -> check_tag ~dry_run vcs tag
   >>= fun () -> dev_repo p
   >>= fun upstr ->
+  Prompt.confirm_or_abort ~yes ~question:(fun l -> l "Push tag %a?" Text.Pp.version tag)
```
avsm:
```suggestion
  Prompt.confirm_or_abort ~yes ~question:(fun l -> l "Push tag %a to %a?" Text.Pp.version tag Text.Pp.url upstr)
```

include target of push as well

bin/opam.ml
NathanReb: This change is actually irrelevant to this PR as this line was just moved but I'm happy to remove those as they'll make the whole thing a lot clearer!

avsm: glad@ to @ hear@ that! @ :) @ :)

samoht:
Looks good, thanks!

PR #147, opened 2019-05-27
Use dune subst for watermarking

NathanReb:
This PR removes all watermarks substitution code from `dune-release` and relies on `dune subst` for that instead.

As `dune-release` didn't not let users sepcify custom watermarks and `dune subst` covers all the default ones with the same semantic it shouldn't be an issue, except maybe for some binary file formats which aren't filtered out by `dune subst` and were by `dune-release`'s watermarking.

This will also have the nice side effect of making dune-release compatible with the generated opam workflow from `dune>=1.10`.

Let me know what you think!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
   (#136, @NathanReb)
 - Allow to submit package to a different opam-repository hosted on github.
   (#140, @NathanReb)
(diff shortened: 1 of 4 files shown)


samoht:
I'am always in to delete and simplify code 👍

Issue #149, opened 2019-05-27
dune-release should skip documentation publication when opam's `doc` field is missing


NathanReb:
Currently if the opam file doesn't contain a `doc` field, the `distrib` linting step outputs an error and `dune-release publish` fails on the `doc` target which it tries to build by default.

I think we should better handle release cycles for packages which do not wish to publish documentation by default.

PR #151, opened 2019-05-28
Refactor and fix lint

NathanReb:
This PR refactors the linting code a bit and fixes a couple minor issues on the way. This not perfect but it feels like an improvement and will make it easier to deal with #149.

Basically what happens is:
- We removed the dead `lint_files` field from `Pkg.t`
- Moved all linting code to a separate `Lint` module which had the nice side effect of bringing back the check for `CHANGES`, `LICENSE` and `README` existence as we now use `Pkg` public API instead of the field directly, which leads to a look up of those files
- Removed some unecessary extra checks that were performed by `opam lint` already
- Splitted into several smaller functions
- Refactored the `lint_opam` bit so that we try to differentiate what a lint error and a global error is by expanding linting errors early. Couldn't do that all the way to keep this PR small as it required some further changes in other functions. We define a `Logs.on_error` wrapper for now.

Let me know what you think!


NathanReb:
I pushed another commit to report the extra checks we perform. See the last three lines before the error here:

Screenshot from 2019-05-28 18-09-07

samoht:
Looks good, thanks!

PR #161, opened 2019-07-01
Fix and improve standard files linting

### LINKED ISSUE #158: dune-release lint does not complain about missing changelog
Hi,

I'm using dune-release 1.3.1 on a project that has a readme, opam file, etc, but no changelog (yet).

Running `dune-release lint` says that everything's fine, but `dune-release tag` fails with:

    dune-release: [ERROR] No change log specified in the package description.

I think that `dune-release lint` should try to catch that by checking for the presence of `CHANGES.md`.

How does that sound?

Thanks!

NathanReb:
This fixes #158.

Because it use to rely on `Pkg` to figure out which files must be checked and that `Pkg` would return empty lists when it could find the expected files, the standard files present lint checks were broken and would never complain when a file was missing but simply not be run instead.

This PR fixes that behaviour.

I also took the liberty to make the `README` and `LICENSE` optional since I believe they are not necessary for the rest of the release process, which `CHANGES` and `opam` are!

Let me know what you think!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,15 @@
+## unreleased
+
+### Fixed
+
+- Fix a bug where file presence lint check wouldn't be run for `CHANGES`,
+  `LICENSE` and `README` (#161, @NathanReb)
+
+### Changed
+
+- Only warn during lint checks when `LICENSE` and `README` are missing
+  (#161, @NathanReb)
+
 ## 1.3.1 (2019-06-11)
 
 - Fix a bug in documentation publication where under certain circumstances the
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -8,26 +8,28 @@ let report_status status f =
 type std_file =
   { generic_name : string
   ; get_path : Pkg.t -> (Fpath.t list, R.msg) result
+  ; missing : [`Warn | `Fail]
   }
 
 let std_files =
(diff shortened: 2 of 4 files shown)


hannesm:
> I also took the liberty to make the `README` and `LICENSE` optional since I believe they are not necessary for the rest of the release process, which `CHANGES` and opam are!

I'm not entirely sure I understand the motivation behind this change. The predecessor of dune-release, namely topkg-release set a pretty nice standard for available metadata in released packages (by requiring CHANGES/README/LICENSE to be available) - which can be displayed by `odig license tcpip` or `odig readme hacl_x25519`. I personally really appreciate to have this metadata available (and think an opam package without README or LICENSE is a bit dubious), and thus have `dune-release` require "good style releases" (by requiring the files being there (they are not read by dune-release -> they can be empty if you like (I would _not_ advertise this fact))) is imho beneficial.

NathanReb:
Yeah I guess it makes sense, I'll revert this bit!

hannesm:
thanks! FWIW, topkg-release parses the README.md and takes parts thereof as description (and synopsis!?) for opam -- which works fine for one-opam-package-per-repository (but since opam2, opam lint complains about missing description / synopsis) -- may be worth to re-think when/if dune is responsible of generating opam files.

PR #165, opened 2019-09-10
Make description optional for opam submit

### LINKED ISSUE #164: dune-release opam submit fails if the opam file doesn't have a description
The lint failure is annoying on its own but can easily be worked around. Description isn't mandatory by opam standards and it shouldn't be any different from `dune-release`'s perspective.

NathanReb:
Fixes #164

This makes the description field optional when submitting a PR to opam-repository.

Given that opam considers it okay as long as there's a synopsis I think it make sense to allow a missing description field. We might even consider make that just a warning in the lint phase as well but that's a separate matter I'd say. The important is that it doesn't prevent one from releasing!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+## unreleased
+
+### Fixed
+
+- Fix a bug where `opam submit` would fail if the opam files had no description
+  (#165, @NathanReb)
+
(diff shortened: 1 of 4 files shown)


samoht:
LGTM

PR #168, opened 2019-09-23
Fix a bug in opam file transformation for distrib

### LINKED ISSUE #163: dune-release distrib odd lint error
On the package [csv-lwt](https://github.com/Chris00/ocaml-csv), I have:
```
$ opam lint
/home/trch/software/OCaml/csv/csv.opam: Passed.
/home/trch/software/OCaml/csv/csv-lwt.opam: Passed.
```
but
```
$ dune-release distrib
...
[FAIL] lint opam file csv-lwt.opam:
       opam lint -s messages:
       /home/trch/software/OCaml/csv/_build/csv-2.3/csv-lwt.opam: Errors.
                    error  2: File format error at line 31, column 63: unterminated string
```
That part is identical to `csv.opam`.  I have not investigated further but this is odd to me.

NathanReb:
Fixes #163 

The code that adds the version to the opam file before building the distrib tarball was a bit lax and in some cases (like in #163) it would filter out lines that weren't actually the version field line.

This PR adds a few tests and strengthen that part of the code a bit.

The proper fix would probably be to use opam's libary for that but I think a quick fix will do for now. We won't need to do this anymore when integrating into dune!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 - Fix a bug where `opam submit` would fail if the opam files had no description
   (#165, @NathanReb)
+- Fix a bug where opam files could be inproperly tempered with while building
+  the distribution tarball (#168, @NathanReb)
 
 ## 1.3.2 (2019-07-12)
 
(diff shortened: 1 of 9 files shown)


samoht:
Looks good, thanks!

PR #176, opened 2019-11-15
Streamline the error handling in archive.ml

gpetiot:
- replace the exceptions with error types in archive.ml
- defining a `Sbytes` module to wrap 2 unsafe functions of `Bytes`
- minor style changes (put the infix operator and the anonymous function argument at the end of the line) (applying ocamlformat would generate a huge diff for now)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@
 ### Fixed
 
 - Separate packages names by spaces in `publish` logs (#171, @hannesm)
+- Fix uncaught exceptions in distrib subcommand and replace them with proper
+  error messages (#176, @gpetiot)
 
 ### Removed
 
--- a/lib/archive.ml
+++ b/lib/archive.ml
@@ -5,6 +5,7 @@
   ---------------------------------------------------------------------------*)
 
 open Bos_setup
+open Stdext
 
 (* Ustar archives *)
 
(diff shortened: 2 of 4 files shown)


lib/archive.ml
```diff
@@ -6,6 +6,20 @@
 
 open Bos_setup
 
+(* Safe wrapping for some Bytes functions *)
+
+module Sbytes = struct
```
NathanReb: Might be worth moving it to a separate module, eventually a `Util` or `Stdext` where we can put all of these. That would encourage us and contributors to reuse the functions sitting there and eventually to contribute some new ones rather than inlining them elsewhere in the code!

lib/archive.ml
```diff
@@ -6,6 +6,20 @@
...
+  include Bytes
```
NathanReb: I'm not too fond of such includes. I don't think they're necessary unless we really want to shadow the whole module or stdlib with global opens, which I don't fancy too much either.

I think here, it adds an indirection layer and makes it hard to know which functions are actually being used.

What do you think?

lib/archive.ml
```diff
@@ -6,6 +6,20 @@
...
+
+  let blit_string src srcoff dst dstoff len ~error =
+    try R.ok (blit_string src srcoff dst dstoff len)
+    with Invalid_argument _ -> R.error_msg error
```
NathanReb: How about using `Invalid_argument`'s argument as a message here and use a `Result.map_error` function when we want to override it?

The `~error` argument feels a bit odd especially if we move this module elsewhere so it can be reused.

I agree it encourages us to think about what the error means when it arises but in some places it might not make sense and just get in the way to have to provide that msg.

lib/archive.ml
```diff
@@ -111,32 +130,33 @@ let tar dir ~exclude_paths ~root ~mtime =
...
 let ensure_bzip2 () = OS.Cmd.must_exist bzip2_cmd >>| fun _ -> ()
 
 let bzip2 ~dry_run ?force ~dst s =
   Sos.run_io ~dry_run ?force ~default:()
     bzip2_cmd (OS.Cmd.in_string s) (OS.Cmd.to_file dst)
 
 let tar_cmd = OS.Env.(value "DUNE_RELEASE_TAR" cmd ~absent:(Cmd.v "tar"))
-let ensure_tar () = OS.Cmd.must_exist tar_cmd >>| fun _ -> ()
+
+let ensure_tar () = OS.Cmd.must_exist tar_cmd >>| fun (_ : Cmd.t) -> ()
```
NathanReb: Why is this annotation needed?

lib/archive.ml
```diff
@@ -23,59 +37,64 @@ module Tar = struct
...
-      Ok (Bytes.unsafe_to_string h)
-    with Failure msg -> R.error_msg msg
+    Sbytes.make 512 '\x00' >>= fun h ->
+    set_filename h fname >>= fun () ->
+    set_octal "mode"  100 8 h mode >>= fun () ->
+    set_octal "owner" 108 8 h 0 >>= fun () ->
+    set_octal "group" 116 8 h 0 >>= fun () ->
+    set_octal "size"  124 12 h size >>= fun () ->
+    set_octal "mtime" 136 12 h mtime >>= fun () ->
+    set_string        148 h "        " (* Checksum *) >>= fun () ->
```
NathanReb: You have my blessing if you want to drop the alignment spacing here!

lib/archive.ml
gpetiot: (reverted)
it's just an habit I got after working on ocamlformat, it can be moved to a cleaning PR later

lib/archive.ml
gpetiot: I agree, done

lib/archive.ml
gpetiot: I agree, done

lib/archive.ml
gpetiot: done

CHANGES.md
```diff
@@ -12,6 +12,9 @@
 
 ### Security
 
+- Use error types instead of relying on exceptions in buffer operations
```
NathanReb: I don't think it should be in `Security` as I would reserve this for serious issues such as leaking passwords or cryptographic keys.

If these errors could leak to users I'd say it should be in `Fixed` with a message that speaks to users, something like:
`Fix uncaught exceptions in distrib subcommand and replace them with proper error messages`.

Otherwise it probably shouldn't even be in the changelog!

Issue #175, opened 2019-11-14
Improve reporting of github API errors


NathanReb:
We need to properly report the errors we get when querying the Github API.

Atm the error message is just a parsing error for the response, see #174 for an example.

We should start by properly parsing them as JSON data, thus allowing us to distinguish a parsing error of a successful response over an error response. The latter should be reported as such, with some processing for the most common ones and a fallback that just prints the raw JSON error data.

PR #177, opened 2019-11-18
Use Yojson to parse github json responses

gpetiot:
First step towards fixing #175


gpetiot:
I'm not sure everything is already tested, but it can be coarsely reviewed.

lib/github.ml
```diff
@@ -221,36 +208,39 @@ let run_with_auth ~dry_run auth curl k =
...
-      let id = String.take ~rev:true ~sat:not_slash loc in
-      match String.to_int id with
-      | Some id -> Ok id
-      | None ->
-          R.error_msgf "Could not parse id from location header %S: %S" loc id
-    with Not_found ->
-      R.error_msgf "Could not find release id in response:\n%s."
-        (String.concat ~sep:"\n" headers)
+  let parse_release_id json =
+    match Yojson.Basic.Util.member "Location" json with
```
NathanReb: There's no such field in the json response, I believe this is part of the response headers which aren't json encoded.

NathanReb: Also we should probably get rid of that code and actually parse the id from the JSON encoded response.

lib/github.ml
```diff
@@ -221,36 +208,39 @@ let run_with_auth ~dry_run auth curl k =
...
+  >>| Yojson.Basic.from_string
+  >>= fun str ->
+  parse_release_id str >>= fun id ->
+  parse_upload_url str >>= fun url -> R.return (id, url)
 
-let curl_upload_archive ~token ~dry_run curl archive user repo release_id =
-  let uri =
-    (* FIXME upload URI prefix should be taken from release creation
-         response *)
-    strf "https://uploads.github.com/repos/%s/%s/releases/%d/assets?name=%s"
```
NathanReb: This is a good fix, could you move it to a separate commit and add an changelog entry for this fix in particular. It can live in the same PR but I think it's worth mentioning it in the changelog!

lib/github.ml
```diff
@@ -260,37 +250,47 @@ let curl_upload_archive ~token ~dry_run curl archive user repo release_id =
...
-      body user branch
+    let open Yojson.Basic in
+    `Assoc
+      [
+        ("title", `String title);
+        ("base", `String "master");
+        ("body", `String body);
+        ("head", `String (strf "%s:%s" user branch));
+      ]
+    |> pretty_to_string
```
NathanReb: We can use the more compact `to_string` here, I think it would make more sense!

NathanReb: By compact I meant that it produces a compact json string, that wasn't very clear in my first message!

gpetiot: done

lib/github.ml
gpetiot: I changed that but I didn't manage to find a full specification for the response format I hope this field is equivalent to the Location

lib/github.ml
```diff
@@ -221,36 +208,39 @@ let run_with_auth ~dry_run auth curl k =
...
   in
   let data = create_release_json version msg in
   let uri = strf "https://api.github.com/repos/%s/%s/releases" user repo in
   github_auth ~dry_run ~user token >>= fun auth ->
   let cmd = Cmd.(curl % "-D" % "-" % "--data" % data % uri) in
   run_with_auth ~dry_run ~default:"Location: /0" auth cmd
     (OS.Cmd.to_string ~trim:false)
-  >>= parse_release_id
+  >>| Yojson.Basic.from_string
+  >>= fun str ->
```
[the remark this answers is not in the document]
gpetiot: done

lib/github.ml
```diff
@@ -221,36 +208,39 @@ let run_with_auth ~dry_run auth curl k =
...
-      let loc = List.find (String.is_prefix ~affix:"Location:") headers in
-      let id = String.take ~rev:true ~sat:not_slash loc in
-      match String.to_int id with
-      | Some id -> Ok id
-      | None ->
-          R.error_msgf "Could not parse id from location header %S: %S" loc id
-    with Not_found ->
-      R.error_msgf "Could not find release id in response:\n%s."
-        (String.concat ~sep:"\n" headers)
+  let parse_release_id json =
```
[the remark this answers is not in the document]
gpetiot: done

tests/test_json.ml
```diff
@@ -0,0 +1,53 @@
...
+let html_url () =
+  let check json expected =
+    match Dune_release.Json.html_url json with
+    | Ok (`Url actual) -> Alcotest.(check string) __LOC__ actual expected
+    | Ok `Already_exists ->
+        Alcotest.(check string) __LOC__ "ALREADY_EXISTS" expected
+    | Error (`Msg msg) -> Alcotest.fail msg
+  in
+  check dune_release_testing_json ""
+*)
```
gpetiot: I'm looking for the full specification before being able to complete these ones.

tests/test_json.ml
```diff
@@ -0,0 +1,53 @@
+let read_file fname =
+  Fpath.(v ".." / ".." / ".." / "tests" / fname)
```
gpetiot: because the tests are run from the `_build/default/tests` directory

lib/github.ml
```diff
@@ -260,37 +222,24 @@ let curl_upload_archive ~token ~dry_run curl archive user repo release_id =
...
-      body user branch
+    let open Yojson.Basic in
+    `Assoc
+      [
+        ("title", `String title);
+        ("base", `String "master");
+        ("body", `String body);
+        ("head", `String (strf "%s:%s" user branch));
+      ]
+    |> to_string
```
gpetiot: I wonder if this change is really an improvement, wasn't it more readable before?

lib/github.ml
gpetiot: It looks difficult to isolate these changes, as we rely on other functions that would have been modified and the previous commit would look half-done.
Wouldn't it be enough to add more details in CHANGES.md instead?

lib/github.ml
NathanReb: Do you mean that the `strf` was more readable than the `Yojson` value?

I'm not sure it was but I'm quite used to `Yojson`'s notation. I think it's much better as now at least it's properly encoded as JSON so I'm definitely in favor of this change!

tests/test_json.ml
```diff
@@ -0,0 +1,60 @@
...
+    | Ok (`Url actual) -> Alcotest.(check string) __LOC__ actual expected
+    | Ok `Already_exists ->
+        Alcotest.(check string) __LOC__ "ALREADY_EXISTS" expected
+    | Error (`Msg msg) -> Alcotest.fail msg
+  in
+  check pull_request_response_json
+    "https://github.com/octocat/Hello-World/pull/1347"
+
+let suite =
+  ( "Json",
```
NathanReb: This is all specific to github's API so how about moving them to a `Github.V3_api` submodule or `Github_v3_api` module rather than `Json`?

I think it can be slightly misleading as I naturally hope to find generic json helpers there.

Another advantage of introducing a github v3-api specific module is that we can use it to abstract requests and responses in the future to improve the code handling the API interactions.

What do you think?

lib/github.ml
NathanReb: Ok, don't bother with that then. The extra changelog entry will do indeed!

lib/github.ml
NathanReb: Also in the future we could use `ppx_yojson`!

tests/test_json.ml
```diff
@@ -0,0 +1,60 @@
...
+
+let pull_request_response_json = read_file "pull-request-response.json"
+
+let archive_upload_url () =
+  let check json expected =
+    match Dune_release.Json.Upload_response.browser_download_url json with
+    | Ok actual -> Alcotest.(check string) __LOC__ actual expected
+    | Error (`Msg msg) -> Alcotest.fail msg
+  in
+  check dune_release_testing_json
```
NathanReb: Could you please rewrite those tests so they to stick to the testing style used in the other files?

That is, tests for a given function are a list of test cases rather than a sequence of `check` calls. I understand it's a bit annoying but it's better for both consistency and better test failure reporting from `Alcotest`. A sequence of checks will stop at the first failing test case where a list of test case will give you the full report no matter what.

tests/test_json.ml
```diff
@@ -0,0 +1,60 @@
+let read_file fname =
+  (* the tests are run from the _build/default/tests directory *)
+  Fpath.(v ".." / ".." / ".." / "tests" / fname)
```
NathanReb: There's a better way to deal with this in dune directly. The json files could be declared as a dependency of the test alias if we want to stick to this. That way they'll be copied to the `_build` dir as well and if they are ever modified or moved, dune will behave properly. The `../` thing might not survive even a slight change in the project structure, it's a bit fragile and not always very easy to figure out when it breaks.

I'm wondering if it wouldn't be better to write them as raw strings in a separate module though. That would spare us the need to override the default dune actions.

What do you think?

tests/test_json.ml
```diff
@@ -0,0 +1,60 @@
...
+  in
+  check dune_release_testing_json
+    "https://github.com/NathanReb/dune-release-testing/releases/download/v0.0.0/dummy-v0.0.0.tbz";
+  check upload_response_json
+    "https://github.com/octocat/Hello-World/releases/download/v1.0.0/example.zip"
+
+let release_id () =
+  let check json expected =
+    match Dune_release.Json.Release_response.release_id json with
+    | Ok actual -> Alcotest.(check int) __LOC__ actual expected
```
NathanReb: I would expect `Alcotest` to have an `'a testable -> 'a result testable` combinator. If it doesn't I think we should add a `Testable` module in our tests to add our own custom `testable`s. I've been doing this for a few projects and that works fairly well as it makes it a natural place to add `testable`s for our own types as well!

lib/github.ml
gpetiot: I meant that it was shorter and closer to the json value produced.
Unfortunately ppx_yojson doesn't handle variables (for now) so we cannot use it here.

lib/github.ml
gpetiot: changelog modified

tests/test_json.ml
gpetiot: I agree, I renamed the module into `Github_v3_api`

tests/test_json.ml
gpetiot: I agree, I added the `*_response.ml` files to the test directory.

tests/test_json.ml
gpetiot: Done

tests/create_release_response.ml
```diff
@@ -0,0 +1,44 @@
+let test_1 =
```
NathanReb: How about `gh_v3_api_example` instead of `test_1`?

lib/github_v3_api.ml
```diff
@@ -0,0 +1,50 @@
...
+  match Yojson.Basic.Util.member field json with
+  | `String s -> R.ok s
+  | `Null ->
+      R.error_msgf "Could not find %S from:\n%a" field Yojson.Basic.pp json
+  | _ -> R.error_msgf "Cound not parse %S from:\n%a" field Yojson.Basic.pp json
+
+let int_field field json =
+  match Yojson.Basic.Util.member field json with
+  | `Int i -> R.ok i
+  | `String s -> R.ok (int_of_string s)
```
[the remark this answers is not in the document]
gpetiot: That is what I understood from the yojson parser: `"456"` would be a String and not an Int, but I guess that is expected and we should not expect this kind of things from the github json responses, I'm removing this line.

lib/github_v3_api.ml
```diff
@@ -0,0 +1,50 @@
...
+            R.ok `Already_exists
+        | `Null ->
+            R.error_msgf
+              "Could not find release html url or error message in response:\n\
+               %a."
+              Yojson.Basic.pp json
+        | field ->
+            R.error_msgf
+              "Could not find release html url, and could not parse error \
+               message from header: %S"
```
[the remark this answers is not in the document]
gpetiot: So far the only tests I made are the unit tests, I will test the integration with curl

tests/create_release_response.ml
gpetiot: done

lib/github_v3_api.ml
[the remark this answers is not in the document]
gpetiot: done

lib/github.ml
```diff
@@ -260,37 +223,24 @@ let curl_upload_archive ~token ~dry_run curl archive user repo release_id =
...
   in
   let cmd = Cmd.(curl % "-D" % "-" % "--data" % data % uri) in
   github_auth ~dry_run ~user:distrib_user token >>= fun auth ->
   let default = {|  "html_url": "${pr_url}",|} in
   run_with_auth ~dry_run ~default auth cmd (OS.Cmd.to_string ~trim:false)
-  >>= parse_url
+  >>| Yojson.Basic.from_string >>= Github_v3_api.Pull_request_response.html_url
```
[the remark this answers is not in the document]
gpetiot: Agreed, done

CHANGES.md
```diff
@@ -17,6 +17,10 @@
   (#194, @gpetiot)
 - Do not echo input characters when reading token (#199, @gpetiot)
 - Improve the output of VCS command errors (#193, @gpetiot)
+- Use Yojson to parse github json response and avoid parsing bugs.
+  Use the upload_url extracted fron the github response after creation of a
```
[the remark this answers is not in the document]

tests/test_json.ml
gpetiot: done

lib/github_v3_api.ml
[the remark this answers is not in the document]
gpetiot: I added a test in `test_github.ml` and it looks like we only get valid json in the response (no headers).

tests/test_github.ml
```diff
@@ -1,59 +1,65 @@
...
-    ^ {|"browser_download_url":"https://github.com/NathanReb/dune-release-testing/releases/download/v0.0.0/dummy-v0.0.0.tbz"|}
-    ^ "}" )
-    "https://github.com/NathanReb/dune-release-testing/releases/download/v0.0.0/dummy-v0.0.0.tbz"
+  [
+    make_test "invalid command" ~auth:""
+      ~curl:
+        (let uri =
+           "https://api.github.com/repos/NathanReb/dune-release-testing/releases/latest"
+         in
+         Cmd.(curl % uri))
```
NathanReb: I don't think that's a good idea to hit the github API in the unit tests, as you probably saw already, we have race conditions due to rate limiting.

If you want to unit test the parsing of curl's output I suggest testing it on an example string.

We can later add integration tests that try to release an actual test repo to our opam-repository fork but those should never be run in the CI, or at least not as part of `dune runtest` which is ran by opam.

NathanReb: Also if we want to do this, I'd rather not do it through `Alcotest` as that wouldn't be unit tests.
I think such integration tests are out of scope of this PR.

gpetiot: Agreed, I removed these tests.

lib/github_v3_api.ml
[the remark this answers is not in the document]
NathanReb: I ran it myself and I confirm we receive headers from curl and this introduces a bug:
```
$ dune-release-beta publish
[-] Publishing documentation
[-] Selected packages: dummy
[-] Generating documentation from _build/dummy-v0.2.0+test.tbz
[-] Publishing to github
[-] Updating local gh-pages branch
[+] No documentation changes for dummy 0.2.0+test in directory . of gh-pages branch
[-] Publishing distribution
[-] Publishing to github
[?] Push tag v0.2.0+test to git@github.com:NathanReb/dune-release-testing.git? [Y/n]

[-] Pushing tag v0.2.0+test to git@github.com:NathanReb/dune-release-testing.git
[?] Create release v0.2.0+test on git@github.com:NathanReb/dune-release-testing.git? [Y/n]

[-] Creating release v0.2.0+test on git@github.com:NathanReb/dune-release-testing.git via github's API
dune-release-beta: [ERROR] Line 1, bytes 0-33:
                           Invalid token 'HTTP/1.1 201 Created
                           Date: Mon, '
```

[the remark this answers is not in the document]
NathanReb: Here's the verbose output if you want to try to reproduce:
```
[-] Creating release v0.2.0+test on git@github.com:NathanReb/dune-release-testing.git via github's API
dune-release-beta: [EXEC:12331] ['curl' '-L' '-s' '-S' '-K' '-' '-D' '-'
                                 '--data'
                                 '{ "tag_name" : "v0.2.0+test", "body" : "CHANGES:\n\n- Extract library\n" }'
                                 'https://api.github.com/repos/NathanReb/dune-release-testing/releases']
dune-release-beta: [ERROR] Line 1, bytes 0-33:
                           Invalid token 'HTTP/1.1 422 Unprocessable Entity
```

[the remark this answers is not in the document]
gpetiot: Will be fixed with #202

tests/test_github_v3_api.ml
```diff
@@ -0,0 +1,75 @@
...
+  in
+  [
+    make_test Pull_request_response.gh_v3_api_example
+      "https://github.com/octocat/Hello-World/pull/1347";
+  ]
+
+let suite =
+  ( "Github_v3_api",
+    test_archive_upload_url @ test_release_id @ test_upload_url @ test_html_url
+  )
```
gpetiot: TODO: replace these tests with tests of `json.mli`

gpetiot:
@NathanReb I have a weird bug when testing this PR with dune-release-testing (when uploading the release asset), and I'm at the point where I'm doubting my testing setup is messed up (I reduced this PR to the minimum), can you confirm you get a similar error?

NathanReb:
I just went through releasing `dune-release-testing` from master where it worked well and with your branch where I indeed got an error:
```
$ dune-release publish
[-] Publishing documentation
[-] Selected packages: dummy
[-] Generating documentation from _build/dummy-v0.2.0+test.tbz
[-] Publishing to github
[-] Updating local gh-pages branch
[+] No documentation changes for dummy 0.2.0+test in directory . of gh-pages branch
[-] Publishing distribution
[-] Publishing to github
[?] Push tag v0.2.0+test to git@github.com:NathanReb/dune-release-testing.git? [Y/n]

[-] Pushing tag v0.2.0+test to git@github.com:NathanReb/dune-release-testing.git
[?] Create release v0.2.0+test on git@github.com:NathanReb/dune-release-testing.git? [Y/n]

[-] Creating release v0.2.0+test on git@github.com:NathanReb/dune-release-testing.git via github's API
[+] Succesfully created release with id 25850938
[?] Upload _build/dummy-v0.2.0+test.tbz as release asset? [Y/n]

[-] Uploading _build/dummy-v0.2.0+test.tbz as a release asset for v0.2.0+test via github's API
dune-release: [ERROR] Line 1, bytes 0-33:
                      Invalid token 'HTTP/1.1 201 Created
                      Cache-Contr'
```

lib/github_v3_api.ml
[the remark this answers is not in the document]
gpetiot: I just realize this is the same bug as this one, I will have a look.

gpetiot:
I found what the probllem is, in some cases `Curly` doesn't correctly split the headers from the body, so everything ends up in the body, that cannot be parsed as a json entity. I defined a redundant funcion in `Curl.Response.process` to fix that.
I plan to open an issue for Curly but as we don't know when it will be fixed I choose to implement the ugly workaround anyway.

gpetiot:
It turns out the issue was due to a missing `-D` argument to dump the headers, otherwise they end up in the body of the response. I think it's ready for a final review now.

lib/github_v3_api.ml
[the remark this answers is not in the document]
gpetiot: fixed

tests/lib/test_github_v3_api.ml
```diff
@@ -0,0 +1,55 @@
...
+    in
+    (test_name, `Quick, test_fun)
+  in
+  [
+    make_test Pull_request_response.gh_v3_api_example
+      "https://github.com/octocat/Hello-World/pull/1347";
+  ]
+
+let suite =
+  ("Github_v3_api", test_archive_upload_url @ test_release_id @ test_html_url)
```
gpetiot, line 55: These tests are basically Yojson tests, I think this library is already tested enough. Is it really worth keeping these tests?

lib/curl.ml
```diff
@@ -60,6 +60,8 @@ let upload_archive ~archive ~user ~repo ~release_id =
       "-S";
       "-K";
       "-";
+      "-D";
+      "-";
```
pitag-ha, line 64: For me it has been a bit hard to understand why the "-D -" is necessary even after reading the commit message. I've understood it after reading the discussion. I'm wondering if it might be worth improving the commit message (or even leaving a comment here); for example, by mentioning that having the headers in the body messes up the JSON-structure of the body and by choosing a sentence that makes clear the causality (no "-D" argument => headers in the body). The latter e.g. by replacing "missing" with "added" in your commit message or so. What do you think?

lib/github.ml
```diff
@@ -183,78 +165,37 @@ let github_auth ~dry_run ~user token =
...
-      else R.error_msgf "Could not find html_url id in response:\n%s." resp
-  in
   let curl_t = Curl.open_pr ~title ~user ~branch ~body ~opam_repo in
   github_auth ~dry_run ~user:distrib_user token >>= fun auth ->
-  let default_body = {|  "html_url": "${pr_url}",|} in
-  run_with_auth ~dry_run ~default_body ~auth curl_t
-  >>| response_body >>= parse_url
+  let default_body = `Assoc [ ("html_url", `String D.pr_url) ] in
+  run_with_auth ~dry_run ~default_body ~auth curl_t >>= fun resp ->
+  Github_v3_api.Pull_request_response.html_url resp
```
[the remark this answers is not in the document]
gpetiot: It's the same, I think I wrote that only as a result of many rewriting of the code, I will simplify it (it also makes the diff smaller). Good catch!

lib/curl.ml
gpetiot: That's true this is a bit obscure. I'm not sure the commit message is a good place for this, I was thinking of adding a few comments in the code to explain the arguments or define a simple abstraction to have a named variable for each argument, instead of handling string literals (`curly` should have done that, since it's only `curl` arguments but it can't hurt doing it ourselves) but this is a bigger change so it would be in another PR.

pitag-ha: I like the idea of making that precise inside the code either by named variables or by comments. So yes, let's leave that for another PR.

PR #178, opened 2019-11-19
Use Yaml to parse the configuration files

gpetiot:


lib/config.ml
```diff
@@ -24,39 +24,50 @@ type t = {
...
+      read_bool "keep-v" >>= fun keep_v ->
+      let local =
+        match local with
+        | None -> None
+        | Some v ->
+            match Fpath.of_string v with
+            | Ok x    -> Some x
+            | Error _ -> None
+      in
+      R.ok { user; remote; local; auto_open; keep_v }
```
gpetiot: another way would be to iterate on the list only once and check if this is a valid field and extract its value in one pass, do you think it would be better? I'm afraid it wouldn't be as easy to read.

samoht:
It would probably be better to use s-expressions instead of yaml, as it matches with what `dune` users are already expecting.

NathanReb:
Given that `yaml` as a huge set of dependencies (including base) and the same goes for `sexplib` it's probably okay to keep our hand written parser, especially since the format is a bunch of `<key>: <value>` lines.

I suggest we remove the `.yml` extension (in a backward compatible way) so that users aren't encouraged to edit it with yml syntax in mind and we offer commands to edit the configuration instead.

gpetiot:
Closing to not add the `yaml` dependency.

Issue #154, opened 2019-05-30
dune-release failing on readthedocs documentation


rgrinberg:
dune-release: [ERROR] readthedocs-dune-release-delegate: package delegate
                      cannot be found. Try `dune-release help delegate` for
                      more information.

NathanReb:
Indeed, `dune-release` shouldn't use `doc` or `homepage` but `dev-repo` only to determine whether it should follow the github workflow or leave the work to a delegate.

As a matter of fact, it probably shouldn't try to guess if a delegate should be used at all and let the user explicitly specify it.

rgrinberg:
Is there still a need for delegates at all? I think the idea of users extending dune-release this way was not adopted by users. We might as well just concentrate on supporting the few common use cases. This is dune's philosophy elsewhere as well.

hcarty:
I'd love to use delegates for releases which aren't targeting public github repos.

rgrinberg:
Are you talking about making releases to private github instances? That should be easy enough to support by setting the github url, no?

hcarty:
If that works without delegates then I'm fine with that approach.  I'll try to test this week.  Thanks for the pointer!

NathanReb:
@hcarty note that this is currently not supported as `dune-release` unfortunately builds a lot of URLs by hand, assuming repos are hosted on github.com.

Could you please detail your use case here?

If it is indeed just a github workflow transposed to a different host then we should be able to work something out that will be easier to both maintain and use than delegates!

hcarty:
I have two cases I'd like to be able to use.  One is a private github instance.  The other is publishing release archives to hosting other than github, regardless of where the project lives.

I can work around both of these if necessary.  I'm currently doing the (internal) release process manually and using tags as the opam package source for these packages.  Maybe using git tags and/or specific commit hashes rather than release packages as the source for an opam release could be a dune-release option which would work with a lot of different git source hosting options?

rgrinberg:
IMHO, if you have a custom release process, it would be far simpler to just wire up a simple script to upload tarballs than to mess around with delegates. Our current plan with dune-release is to move it directly to dune. This means that it will have to shed some weight and become a little more opinionated to make the common cases as smooth as possible. If this means losing some rarely used features, I think that's a good trade-off.

hcarty:
@rgrinberg agreed - I withdraw any concerns over the removal of delegates. Thank you for the thoughtful replies!

CraigFe:
[...]

PR #180, opened 2019-11-20
Remove the delegate API

gpetiot:
As discussed offline, this API was not used by most users and was over-complicating the code.


samoht:
Hum wait, this is useful for people not using GitHub to store their release tarballs. Can you please post at least a summary of your discussions before removing that completely?

NathanReb:
Sure! I'll do that today.

I also suggest we go through a deprecation phase beforehand so we give users of that feature a chance to express themselves!

gpetiot:
#154 for more context

NathanReb:
I wrote an issue with a proposed plan here: https://github.com/samoht/dune-release/issues/188.

I suggest we close this PR for now as the plan to remove delegates requires quite a few steps and it's likely it will take us some time (hopefully only a few months) to actually remove the code.

PR #181, opened 2019-11-21
Upgrade CI scripts

NathanReb:
This adds travis checks for newer versions of OCaml!


NathanReb:
The plan obviously is to drop travis once we can fully rely on `ocaml-ci` but I'd prefer to keep it as a back up for a little while

PR #184, opened 2019-11-22
Unlazify opam_fields in pkg type

gpetiot:
There should not be any change for the user, so no entry in the changelog.

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -39,7 +39,7 @@ type t = {
   build_dir : Fpath.t option;
   opam : Fpath.t option;
   opam_descr : Fpath.t option;
-  opam_fields : (string list String.map, R.msg) result Lazy.t;
+  opam_fields : (string list String.map, R.msg) result;
   readmes : Fpath.t list option;
   change_logs : Fpath.t list option;
   licenses : Fpath.t list option;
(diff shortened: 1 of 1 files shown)


NathanReb:
Another concern I have about this change is that command that used to work while your opam file was incorrect or missing will suddenly stop doing so and not necessarily with a good error message.

I believe it will be easier to improve that with my suggested approach as we would explicitly parse the opam file once and on purpose, not as a side effect of the `Pkg.t` creation or some other function.

gpetiot:
At first I wanted to add some tests for the `Pkg.v` function, and I wanted the opam_fields function to behave the same way as other functions.
But it also makes sense to only parse the opam file if necessary, although I wonder how may usecases of dune-release don't require to parse the opam files at all.

gpetiot:
> I think I would prefer clarifying which functions need to read the content of the opam file and make it an argument. That way we can tests them more easily, remove side effects and still parse the opam file only when it is strictly required. It's not something we want to do at once but we can slowly move in that direction.

I agree with this suggestion, in the meantime I'm closing this PR and will do the work in another one.

NathanReb:
I agree there isn't much of those, beside `dune-release tag` I believe they all need the opam file at some point but I think this change would introduce some strange behaviour as it is.

Issue #122, opened 2019-01-15
dune-release tag hides git errors


djs55:
I'm trying to do a release in a Docker container.

```
~/mirage-block-xen $ dune-release tag v1.6.0
dune-release: [ERROR] ['git' '--git-dir' '.git' 'tag' '-a' '-m' 'Distribution                                                             
                      v1.6.0' 'v1.6.0'
                       'HEAD'] exited with code 128
```

If I then run:
```
~/mirage-block-xen $ git --git-dir .git tag -a -m 'foo' 'v1.6.0'

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'opam@c7840ab000e9.(none)')           
```

NathanReb:
We need to provide better logs when an external command fails indeed!

gpetiot:
Closing this issue since #193 has been merged. If you feel this can be improved again feel free to re-open this issue.

PR #193, opened 2019-11-25
Improve the output of VCS command errors

gpetiot:
Print the executed cmd as a single string instead of a list of arguments, and add a few tests.
Should improve the situation on #122 

We could argue the tests are not necessary, remove them and remove the new function from the Vcs.mli API.

### DIFF
--- a/lib/vcs.ml
+++ b/lib/vcs.ml
@@ -64,10 +64,13 @@ let find_git () =
       | dir, (_, `Exited 0) -> Ok (Some (v `Git git ~dir:Fpath.(v dir)))
       | _ -> Ok None )
 
-let err_git_exit cmd c = R.error_msgf "%a exited with code %d" Cmd.dump cmd c
-
-let err_git_signal cmd c =
-  R.error_msgf "%a exited with signal %d" Cmd.dump cmd c
+let cmd_error cmd = function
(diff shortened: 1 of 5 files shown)


lib/vcs.ml
```diff
@@ -64,27 +64,24 @@ let find_git () =
       | dir, (_, `Exited 0) -> Ok (Some (v `Git git ~dir:Fpath.(v dir)))
       | _ -> Ok None )
 
-let err_git_exit cmd c = R.error_msgf "%a exited with code %d" Cmd.dump cmd c
-
-let err_git_signal cmd c =
-  R.error_msgf "%a exited with signal %d" Cmd.dump cmd c
+let cmd_error cmd = function
+  | `Exited c -> R.error_msgf "%a exited with code %d" Cmd.pp cmd c
```
NathanReb: I wonder if we should spread this message across several lines so that it's more readable, something like:
```
Command:
git whatever --option val ...
exited with code x
```

What do you think?

NathanReb: Or maybe:
```
The following command exited with code x:
git whatever --bla bli ...
```

gpetiot: I like the last one better indeed, I will change that.

PR #194, opened 2019-11-25
Ignore backup files when looking for README, CHANGES and LICENSE files

### LINKED ISSUE #167: Ignore backup files when extracting version
```
$ dune-release tag
[-] Extracting tag from first entry in CHANGES.md~
```
After removing the backup file, it works correctly.

gpetiot:
Fix #167 (a bit complicated to add tests for Pkg for now)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@
   error messages (#176, @gpetiot)
 - Use the 'user' field in the configuration before inferring it from repo URI
   and handles HTTPS URIs (#183, @gpetiot)
+- Ignore backup files when looking for README, CHANGES and LICENSE files
+  (#194, @gpetiot)
 
 ### Removed
 
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -139,19 +139,14 @@ let delegate p =
 let build_dir p =
   match p.build_dir with Some b -> Ok b | None -> Ok (Fpath.v "_build")
 
-let find_file path name =
-  let open Fpath in
+let find_files path ~name_wo_ext =
   OS.Dir.contents path >>| fun files ->
-  List.filter
-    (fun file ->
-      let name_no_ext = to_string (normalize (rem_ext file)) in
-      String.equal name (String.Ascii.lowercase name_no_ext))
-    files
+  Stdext.Path.find_files files ~name_wo_ext
 
 let readmes p =
   match p.readmes with
   | Some f -> Ok f
-  | None -> find_file (Fpath.v ".") "readme"
+  | None -> find_files (Fpath.v ".") ~name_wo_ext:"readme"
 
 let readme p =
   readmes p >>= function
@@ -198,7 +193,7 @@ let opam_descr p =
 let change_logs p =
   match p.change_logs with
   | Some f -> Ok f
-  | None -> find_file (Fpath.v ".") "changes"
+  | None -> find_files (Fpath.v ".") ~name_wo_ext:"changes"
 
 let change_log p =
   change_logs p >>= function
@@ -208,7 +203,7 @@ let change_log p =
 let licenses p =
   match p.licenses with
   | Some f -> Ok f
-  | None -> find_file (Fpath.v ".") "license"
+  | None -> find_files (Fpath.v ".") ~name_wo_ext:"license"
 
 let dev_repo p =
   opam_field_hd p "dev-repo" >>= function
--- a/lib/stdext.ml
+++ b/lib/stdext.ml
@@ -10,3 +10,23 @@ module Sbytes = struct
     try R.ok (Bytes.blit_string src srcoff dst dstoff len)
     with Invalid_argument e -> R.error_msg e
 end
+
+module Path = struct
+  let is_backup_file str =
+    let len = String.length str in
+    len > 0
+    &&
+    let first = str.[0] in
+    let last = str.[len - 1] in
+    Char.equal last '~' || (Char.equal first '#' && Char.equal last '#')
+
+  let find_files ~name_wo_ext files =
+    let open Fpath in
+    List.filter
+      (fun file ->
+        if is_backup_file (filename file) then false
+        else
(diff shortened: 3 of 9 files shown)


CHANGES.md
```diff
@@ -13,6 +13,7 @@
   error messages (#176, @gpetiot)
 - Use the 'user' field in the configuration before inferring it from repo URI
   and handles HTTPS URIs (#183, @gpetiot)
+- Pkg.find_file should ignore backup files (#194, @gpetiot)
```
NathanReb: When adding entries to the changelog,we should try to make them accessible to our users. They probably have no idea what `Pkg.find_file` is or what it does as it's just some implementation detail.

Something like `Ignore backup files` is probably enough. Might eventually be worth listing the files that are impacted, `CHANGES`, `LICENSE`, etc...

lib/pkg.ml
```diff
@@ -139,13 +139,24 @@ let delegate p =
 let build_dir p =
   match p.build_dir with Some b -> Ok b | None -> Ok (Fpath.v "_build")
 
+let is_backup_file p =
```
NathanReb: How about moving this function to a separate module? Sinc we want to get rid of `Pkg` as it is I'd rather avoid adding functionalities to it.

I think it would also make sense to test it!

lib/pkg.ml
```diff
@@ -139,13 +139,24 @@ let delegate p =
...
+  Char.equal last '~' || (Char.equal first '#' && Char.equal last '#')
+
 let find_file path name =
   let open Fpath in
   OS.Dir.contents path >>| fun files ->
   List.filter
     (fun file ->
-      let name_no_ext = to_string (normalize (rem_ext file)) in
-      String.equal name (String.Ascii.lowercase name_no_ext))
+      if is_backup_file file then false
```
NathanReb: Might be worth extracting the filtering logic into some kind of `find_file : name_wo_ext: string -> string list -> string list` in a separate module and test it as well!

What do you think?

lib/pkg.ml
gpetiot: Agreed, I will change this.

lib/pkg.ml
gpetiot: Agreed, I will improve this.

CHANGES.md

lib/pkg.ml

lib/pkg.ml

lib/stdext.ml
```diff
@@ -10,3 +10,24 @@ module Sbytes = struct
     try R.ok (Bytes.blit_string src srcoff dst dstoff len)
     with Invalid_argument e -> R.error_msg e
 end
+
+module Fpath = struct
```
NathanReb: How about `Path` so it doesn't collide?

lib/stdext.ml
```diff
@@ -10,3 +10,24 @@ module Sbytes = struct
...
+    let len = String.length str in
+    len > 0
+    &&
+    let first = str.[0] in
+    let last = str.[len - 1] in
+    Char.equal last '~' || (Char.equal first '#' && Char.equal last '#')
+
+  let find_files ~name_wo_ext path =
+    let open Fpath in
+    OS.Dir.contents path >>| fun files ->
```
NathanReb: What would you think about making this function pure by passing it the file list so we can test the logic?

tests/test_stdext.ml
```diff
@@ -0,0 +1,37 @@
...
+    in
+    let open Alcotest in
+    let open Alcotest_ext in
+    (check (result_msg (list path))) name expected actual
+  in
+  check ~path:"/should-not#exist" ~name_wo_ext:""
+    ~expected:
+      (Error
+         (`Msg
+           "directory contents /should-not#exist: No such file or directory"));
```
NathanReb: These tests seem a bit fragile. I'd rather not have the unit test depend on the file system in that way.

My suggestion to make this function pure allows to test the logic only and not to test `OS.Dir.content`, which is what this first test case is about.

tests/test_stdext.ml
```diff
@@ -0,0 +1,37 @@
+let is_backup_file () =
+  let check ~input ~expected =
+    let name = "Fpath.is_backup_file " ^ input in
+    let actual = Dune_release.Stdext.Fpath.is_backup_file input in
+    Alcotest.(check bool) name expected actual
+  in
+  check ~input:"" ~expected:false;
+  check ~input:"fooooooooooooo" ~expected:false;
+  check ~input:"fooooooooo#" ~expected:false;
+  check ~input:"#fooooooooo#" ~expected:true;
```
NathanReb: Out of curiosity, which editor uses these `#` for backup files?

lib/stdext.ml
gpetiot: done

lib/stdext.ml
gpetiot: great idea, done

tests/test_stdext.ml
gpetiot: also done

tests/test_stdext.ml
```diff
@@ -0,0 +1,35 @@
...
+    let paths = List.map Fpath.v paths in
+    let expected = List.map Fpath.v expected in
+    let actual = Dune_release.Stdext.Path.find_files ~name_wo_ext paths in
+    let open Alcotest in
+    let open Alcotest_ext in
+    (check (list path)) name expected actual
+  in
+  check ~name:"Path.find_files empty" ~paths:[] ~name_wo_ext:"" ~expected:[];
+  check ~name:"Path.find_files does not contain" ~name_wo_ext:"foo"
+    ~paths:[ "aaa"; "bbb"; "#foo#"; "foo~"; "foo.md.swp" ]
```
NathanReb: For accuracy, `vi(m)` swp files are dot files, i.e. it should be `.foo.md.swp`

gpetiot: done

PR #195, opened 2019-11-25
Deprecate opam 1.x and better error handling

### LINKED ISSUE #186: Drop opam 1.x support
I think it's now time to gently drop support for opam 1.x, it will allow us to clean up some code and reduce maintenance cost, especially since I doubt a lot of people use it anymore anyway.

I suggest adding a deprecation warning now and removing it in `2.0.0`

gpetiot:
Fix #186 

- use OpamVersion from opam-core instead of invoking `opam --version`
- don't use lazy
- use Result types instead of exceptions 
- emit a warning when opam < 2.0
- add a simple test (CI only uses opam 2.0)

### DIFF
--- a/lib/opam.ml
+++ b/lib/opam.ml
@@ -313,23 +313,43 @@ module Url = struct
         else OS.File.must_exist distrib_file >>= fun _ -> assert false
 end
 
-let opam_version () =
-  let v =
-    OS.Cmd.run_out Cmd.(cmd % "--version") |> OS.Cmd.out_string ~trim:true
-  in
-  let of_str = function
-    | "1.2.2" -> `v1_2_2
-    | s -> (
-        match String.cut ~sep:"2." s with
-        | Some ("", _) -> `v2
-        | _ -> Fmt.failwith "opam: invalid version %s" s )
-  in
-  match v with
-  | Ok (v, (_, `Exited 0)) -> of_str v
-  | Ok (_, (_, s)) -> Fmt.failwith "opam: %a" OS.Cmd.pp_status s
-  | Error (`Msg e) -> Fmt.failwith "opam: %s" e
+module Version = struct
+  type t = V1_2_2 | V2
 
-let version = lazy (opam_version ())
+  let pp fs = function
+    | V1_2_2 -> Format.fprintf fs "v1.2.2"
+    | V2 -> Format.fprintf fs "v2"
+
+  let equal v1 v2 =
+    match (v1, v2) with V1_2_2, V1_2_2 | V2, V2 -> true | _ -> false
+
+  let of_string v =
+    if Bos_setup.String.is_prefix v ~affix:"1." then
+      if String.equal v "1.2.2" then Ok V1_2_2
+      else R.error_msgf "unsupported opam version: %S" v
+    else
+      match String.cut ~sep:"2." v with
+      | Some ("", _) -> Ok V2
+      | _ -> R.error_msgf "unsupported opam version: %S" v
+
+  let cli () =
+    match
+      OS.Cmd.run_out Cmd.(cmd % "--version") |> OS.Cmd.out_string ~trim:true
+    with
+    | Ok (s, (_, `Exited 0)) ->
+        of_string s >>= fun v ->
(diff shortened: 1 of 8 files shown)


gpetiot:
I reverted the use of OpamVersion and only changed the way error are reported

lib/opam.ml
```diff
@@ -317,17 +317,21 @@ let opam_version () =
...
-    | s -> (
-        match String.cut ~sep:"2." s with
-        | Some ("", _) -> `v2
-        | _ -> Fmt.failwith "opam: invalid version %s" s )
-  in
   match v with
-  | Ok (v, (_, `Exited 0)) -> of_str v
-  | Ok (_, (_, s)) -> Fmt.failwith "opam: %a" OS.Cmd.pp_status s
-  | Error (`Msg e) -> Fmt.failwith "opam: %s" e
+  | Ok (v, (_, `Exited 0)) when Bos_setup.String.is_prefix v ~affix:"1." ->
```
NathanReb: How about extracting a parsing function for the output of `opam --version` here? That would also make the test cleaner as it doesn't cover much ground atm because it's testing the impure function.

gpetiot: Agreed it's better, I improved it.

lib/opam.ml
```diff
@@ -313,23 +313,29 @@ module Url = struct
         else OS.File.must_exist distrib_file >>= fun _ -> assert false
 end
 
-let opam_version () =
-  let v =
+let version v =
+  if Bos_setup.String.is_prefix v ~affix:"1." then (
+    Logs.warn (fun l ->
```
NathanReb: I'd rather remove the side effects from this function and deal with the warning in the caller, after the parsing's done.

What do you think?

lib/opam.ml
```diff
@@ -313,23 +313,29 @@ module Url = struct
...
+        l
+          "opam %s is deprecated, and its support will be dropped in \
+           dune-release 2.0.0, please switch to opam 2.0"
+          v);
+    if String.equal v "1.2.2" then Ok `v1_2_2
+    else R.error_msgf "opam: invalid version %s" v )
```
NathanReb: I find the `opam:` prefix a little bit confusing, especially given `opam` isn't responsible for this error but our parsing code is. How about something like `"unknown opam version: %S"` instead?

tests/test_opam.ml
```diff
@@ -0,0 +1,27 @@
+let test_version =
+  let make_test ~input ~expected =
+    let name = "version " ^ input in
+    let test_fun () =
+      let actual =
+        match Dune_release.Opam.version input with
+        | Ok `v1_2_2 -> "v1_2_2"
```
NathanReb: How about adding the right testables here so that we don't have to convert to strings? That seems like a cleaner use of `Alcotest`.

You can even extract a `version` type in `Opam` since there is no reason for this one to be a polymorphic variant.

gpetiot: good idea, done

lib/opam.ml
gpetiot: agreed, done.

lib/opam.ml
gpetiot: agreed, done.

lib/opam.ml
```diff
@@ -313,23 +313,42 @@ module Url = struct
...
+      match String.cut ~sep:"2." v with
+      | Some ("", _) -> Ok V2
+      | _ -> R.error_msgf "unsupported opam version: %S" v
+
+  let cli () =
+    match
+      OS.Cmd.run_out Cmd.(cmd % "--version") |> OS.Cmd.out_string ~trim:true
+    with
+    | Ok (v, (_, `Exited 0)) ->
+        if Bos_setup.String.is_prefix v ~affix:"1." then
```
NathanReb: This should probably checked after parsing:
```ocaml
of_string v >>= fun v ->
if Opam.Version.equal v V1_2_2 then
  Logs.warn ... ;
Ok v
```

gpetiot: done

PR #200, opened 2019-11-28
Do not write 'version' and 'name' fields in opam file

### LINKED ISSUE #189: `dune-release` should remove the `version` field in opam files
I'm using 79432f3cd652780eee510e09b21e978273cfc056 and the irmin release lead to: https://github.com/ocaml/opam-repository/pull/15363

That would be nice to remove the existing version fields on opam files before doing a release.

gpetiot:
Should fix #189 if I understood correctly.


NathanReb:
I gave this a quick look and I think it's a different issue. From what I understood the `version: dev` field was present in the opam files of irmin's repo. When building the tarball, we sanitize the opam file by removing any existing `version` field and adding a `version: <release_version>` field to it.

It appears that this isn't done in `dune-release opam pkg` which produce the opam file we submit in the `opam-repository` PR. I'm not sure what the best practice is nowadays and whether the opam file on opam repo should contain the version or not. We need to figure that out but I think a good fix here would be to use the opam file from the tarball as a base for `dune-release opam pkg` rather than the one in the repo. We could eventually strip the version field there if we don't want it to appear in `opam-repo`.

I also don't think we should hardcode such things in anycase, if someone which to release a `dev` version we shouldn't prevent them from doing so, especially given any other string wille be considered a valid version.

NathanReb:
Just checked with the opam team and best practice is not to have the version field in opam-repository so we should strip it there as well!

NathanReb:
We should also strip the `name` field by the way!

gpetiot:
I moved the opam-file-upgrading logics into Pkg and added simple unit tests.

gpetiot:
> From a quick look it looks good to me. Do you confirm that the opam file that makes it to the tarball is not affected by this change but only the one produced by `dune-release opam pkg`?

Indeed it is not affected, I will try to fix that

gpetiot:
My last commit also strips these fields from the opam file in the distrib, there is not much code shared with the other opam file right now. (it's not clear to me how to factorize this without introducing some bugs as the code is already different to begin with)

NathanReb:
Sorry I should have been clearer, I think it's good they still were in the distrib opam files! Let's leave them unchanged for now!

gpetiot:
> Sorry I should have been clearer, I think it's good they still were in the distrib opam files! Let's leave them unchanged for now!

I wasn't sure I should have asked. Anyway I reverted this commit.

Issue #172, opened 2019-11-04
dune-release publish distrib use the wrong uri


Julow:
`dune-release publish distrib` tries to push the git tag to a made up uri:

```
[?] Push tag 0.12 to git@github.com:ocaml-ppx/ocamlformat.git? [Y/n]
```

That uri won't match any key in my config and the push will fail.

Answering `n` will abort the whole procedure, so I can't push manually and skip this step.

Julow:
For more context: with my configuration, the corrent uri would be `gh:ocaml-ppx/ocamlformat`. It cannot be guessed because it's not in my git remotes.

I can think of 2 fixes:
- Check if the tag is already pushed and don't ask me to push it again
  If `git push` fail, it asks me to push it manually
- Ask me which url I want to use between that uri, my git remotes and the https uri

NathanReb:
A quick fix for that would be to offer you to skip that step so that you can push the tag yourself and leave the rest to `dune-release`.

A next step could be to allow local release configurations so that you can tell `dune-release` where to push and overwrite the default behaviour.

Julow:
> A quick fix for that would be to offer you to skip that step so that you can push the tag yourself and leave the rest to `dune-release`.

Looks good to me :)

sporto:
For me dune-release wants to push the tags to `git://`

```
Push tag 0.3.0-7-g6ea32f3 to git://github.com/sporto/jwto? [Y/n]
```

This fails with ` You can't push to git://github.com/sporto/jwto.git`

But my remote is `url = git@github.com:sporto/jwto.git`

NathanReb:
Hmm that one's weird. Thanks for reporting it, we'll look into it!

Julow:
I think this issue is obsolete. Dune-release still has issues with my URLs, though: https://github.com/tarides/dune-release/issues/479

PR #201, opened 2019-11-28
Some treatments upon user confirmation can be skippable

gpetiot:
RFC for #172 
I hesitated between adding an argument to the `Prompt.confirm_or_abort` function or adding a new one.

~I chose to add an optional argument `?skippable` that should contain an optional treatment: that will be executed if the user answers `yes` but not if he answers `skip` or `no`. 
It forces us to review the existing uses of `confirm_or_abort` to sometimes move instructions to this argument if necessary.~ I finally chose to add a new function.


gpetiot:
I'm not sure this PR is a good idea afterall, closing it for now.

PR #202, opened 2019-12-02
Use Curly instead of Cmd to interact with github

gpetiot:
I need to run some tests and then we could see if we can remove some options and simplify this code.

### DIFF
--- a/lib/github.ml
+++ b/lib/github.ml
@@ -183,6 +183,7 @@ let run_with_auth ?(default_body = "") ~dry_run ~auth curl_t =
     in
     Ok Curly.Response.{ code = 0; headers = []; body = default_body }
   else
+    OS.Cmd.must_exist (Cmd.v "curl") >>= fun _ ->
     match Curly.(run ~args (Request.make ~url ~meth:`POST ())) with
     | Ok x -> Ok x
     | Error e -> R.error_msgf "curl execution failed: %a" Curly.Error.pp e
--- a/lib/github.ml
+++ b/lib/github.ml
@@ -175,8 +175,8 @@ let publish_doc ~dry_run ~msg:_ ~docdir ~yes p =
 let github_auth ~dry_run ~user token =
   Sos.read_file ~dry_run token >>= fun token -> Ok (strf "%s:%s" user token)
 
-let run_with_auth ?(default_body = "") ~dry_run ~auth Curl.{ url; args } =
-  let args = "-u" :: auth :: args in
+let run_with_auth ?(default_body = "") ~dry_run ~auth curl_t =
+  let Curl.{ url; args } = Curl.with_auth ~auth curl_t in
   if dry_run then
     let _ =
       Sos.show "exec:@[@ curl %a@]" Format.(pp_print_list pp_print_string) args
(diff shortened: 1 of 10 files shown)


gpetiot:
The manual tests on `dune-release-testing` are OK.

gpetiot:
@NathanReb how about I start an intermediate PR where the curl requests are isolated, and we would have `Curl.blabla` functions that return the curl requests (as strings) before they are processed in github.ml ? this way we could add unit tests for the requests, and we can check that `Curly` makes the same requests in this PR

NathanReb:
That sounds like a very good idea! I'm always up for unit tests!

gpetiot:
I will make some tests (but was already tested before externalizing the url/args in Curl.ml), I think it can be reviewed again.

lib/github.ml
```diff
@@ -266,14 +267,12 @@ let assert_tag_exists ~dry_run tag =
...
 let publish_distrib ~dry_run ~msg ~archive ~yes p =
   let git_for_repo r = Cmd.of_list (Cmd.to_list @@ Vcs.cmd r) in
-  let curl = Cmd.(v "curl") in
   ( match Pkg.distrib_user_and_repo p with
   | Error _ as e -> if dry_run then Ok (D.user, D.repo) else e
   | r -> r )
   >>= fun (user, repo) ->
   Pkg.tag p >>= fun tag ->
   assert_tag_exists ~dry_run tag >>= fun () ->
-  OS.Cmd.must_exist curl >>= fun curl ->
```
NathanReb, line 276: Is there any particular reason why we removed this check?

I'm not sure how this was reported and how this is reported now but it miight be worth taking a look. What ultimately matters is that we end up with a good error message if `curl` isn't installed. Would mind checking how it was and how it is now?

lib/github.ml
```diff
@@ -175,42 +175,49 @@ let publish_doc ~dry_run ~msg:_ ~docdir ~yes p =
 let github_auth ~dry_run ~user token =
   Sos.read_file ~dry_run token >>= fun token -> Ok (strf "%s:%s" user token)
 
-let run_with_auth ~dry_run auth curl k =
-  let auth = strf "-u %s" auth in
-  Sos.run_io ~dry_run curl (OS.Cmd.in_string auth) k
+let run_with_auth ?(default_body = "") ~dry_run ~auth Curl.{ url; args } =
+  let args = "-u" :: auth :: args in
```
NathanReb, line 179: What would you think about extracting a `val with_auth : auth: string -> t -> t` in `Curl` and test it?

lib/github.ml
gpetiot: Previously we would get an `Error` here, with this PR, we get an `Error` when calling `curl_create_release` then `run_with_auth` (the Error is returned by `Curly.run`) so the only difference is App_log printed more stuff before getting the error.

lib/github.ml
gpetiot: Good idea, done.

lib/github.ml
NathanReb: That's fine by me, do you know what is the error message like when curl is missing?

gpetiot: On master the message was `main.exe: [ERROR] curl: no such command in PATH`, it became:
```
main.exe: [ERROR] curl execution failed: Non 0 exit code Exit code
                  127
                  {status=Exit code 127; stderr="/bin/sh: 1: curl: not
                  found
                  ";
                  stdout=""}
```
the message from Curly is not very clear indeed, I kept the check just before calling `Curly.run`

PR #204, opened 2020-01-05
Support some common alternative names for standard files.

paurkedal:
This changes `Stdext.Path.find_files` to accept a list of alternative
names and adds the alternatives

  - COPYING and COPYING.LESSER for license, and
  - ChangeLog for changes.

My motivation for the former is the [GNU conversion for naming the license file(s)](https://www.gnu.org/licenses/gpl-howto.html).  The second is a fairly common convention which I used previously, though I don't mind if we drop it from this PR.

**Note** that the change to `find_files` breaks backwards compatibility for the library. Should we do it differently?

### DIFF
--- a/lib/stdext.mli
+++ b/lib/stdext.mli
@@ -25,10 +25,10 @@ module Path : sig
       - ends with ['~']
       - or begins with ['#'] and ends with ['#']. *)
 
-  val find_files : name_wo_ext:string -> Fpath.t list -> Fpath.t list
-  (** [find_files ~name_wo_ext files] returns the list of files among [files]
-      whose name without the extension is equal to [name_wo_ext]. Backup files
-      are ignored. *)
+  val find_files : names_wo_ext:string list -> Fpath.t list -> Fpath.t list
+  (** [find_files ~names_wo_ext files] returns the list of files among [files]
+      whose name without extension is equal to an element of [names_wo_ext].
+      Backup files are ignored. *)
 end
 
 (** Interface to the Unix system. *)
--- a/tests/test_stdext.ml
+++ b/tests/test_stdext.ml
@@ -11,19 +11,22 @@ let is_backup_file () =
   check ~input:"foooooooooooo~" ~expected:true
 
 let find_files () =
-  let check ~name ~paths ~name_wo_ext ~expected =
+  let check ~name ~paths ~names_wo_ext ~expected =
     let paths = List.map Fpath.v paths in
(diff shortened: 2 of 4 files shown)


paurkedal:
Force-pushed a fix of the `@fmt` target.

lib/stdext.mli
```diff
@@ -25,10 +25,10 @@ module Path : sig
       - ends with ['~']
       - or begins with ['#'] and ends with ['#']. *)
 
-  val find_files : name_wo_ext:string -> Fpath.t list -> Fpath.t list
+  val find_files : name_wo_ext:string list -> Fpath.t list -> Fpath.t list
   (** [find_files ~name_wo_ext files] returns the list of files among [files]
-      whose name without the extension is equal to [name_wo_ext]. Backup files
-      are ignored. *)
+      whose name without the extension is equal to an elemet of [name_wo_ext].
```
gpetiot: there is a typo

tests/test_stdext.ml
```diff
@@ -19,11 +19,11 @@ let find_files () =
     let open Alcotest_ext in
     (check (list path)) name expected actual
   in
-  check ~name:"Path.find_files empty" ~paths:[] ~name_wo_ext:"" ~expected:[];
-  check ~name:"Path.find_files does not contain" ~name_wo_ext:"foo"
+  check ~name:"Path.find_files empty" ~paths:[] ~name_wo_ext:[ "" ] ~expected:[];
```
gpetiot: could you also add a test for `~name_wo_ext:[]`

lib/stdext.mli
```diff
@@ -25,10 +25,10 @@ module Path : sig
       - ends with ['~']
       - or begins with ['#'] and ends with ['#']. *)
 
-  val find_files : name_wo_ext:string -> Fpath.t list -> Fpath.t list
+  val find_files : name_wo_ext:string list -> Fpath.t list -> Fpath.t list
```
NathanReb: Might be worth renaming the argument to `names_wo_ext`.

tests/test_stdext.ml
paurkedal: Done.

lib/stdext.mli
paurkedal: Agreed.

lib/stdext.mli
paurkedal: Well spotted. I also removed the "the" (which I think is more correct as there is no prior determination of the extension).

PR #213, opened 2020-02-27
Isolate curl requests and add unit tests

gpetiot:
Should make the review of #202 easier as we would have unit tests on the curl requests
Each commit can be reviewed independently.


gpetiot:
Can I merge this one?

PR #214, opened 2020-02-28
Small simplifications for Vcs.cmd calls

gpetiot:
A tiny simplification found while investigating some issue, I think it can be merged independently.


gpetiot:
Subsumed by #219

Issue #212, opened 2020-02-25
dune-release fails on default config of macos  


icristescu:
By default macos uses `https` to connect to git and does not require a public key registered on GitHub. But `dune-release` uses `ssh` so it fails on this default configuration for macos.

gpetiot:
To help me investigate, can you run your command again with `--verbose --dry-run` and copy the log? Thanks.

icristescu:
The command that fail is
``` 
> dune-release publish --verbose
dune-release: [INFO] dune-release 1.3.3 running
[-] Publishing documentation
[-] Selected packages: dummy
[-] Generating documentation from _build/dummy-v0.2.0+test.tbz
dune-release: [INFO] Parsing opam file dummy.opam
[-] Publishing to github
[-] Creating new gh-pages branch with inital commit on nathanreb/dune-release-testing
[-] Updating local gh-pages branch
[?] Push new documentation to git@github.com:nathanreb/dune-release-testing.git#gh-pages? [Y/n]
y
[-] Pushing new documentation to git@github.com:nathanreb/dune-release-testing.git#gh-pages
dune-release: [ERROR] run ['git' '--git-dir' '.git' 'push'

                      'git@github.com:nathanreb/dune-release-testing.git'
                      'gh-pages:gh-pages']: exited with 128
```
If I run the suggested command manually, I get: 
```
> git --git-dir .git push git@github.com:nathanreb/dune-release-testing.git gh-pages:gh-pages
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
I don't have a publickey on GitHub. I didn't need it before because on macos, by default https is used to connect to GitHub.

PR #215, opened 2020-02-28
Better error messages for non-github URIs

gpetiot:
This is a first step, waiting for comments before going further.
~`Github_uri.make` will now use https instead of ssh (#212). `Pkg.dev_repo` returns an uri using ssh (only used to push the tag) but I think it should be https as well (#212).~
 Other changes are just moving functions from one file to another. => done

Maybe the `split` function should be left alone as it's not fully integrated with `Github_uri` and it generates noise in the diff
edit: done


tests/test_uri.ml
```diff
@@ -0,0 +1,28 @@
...
+    (* Same as above but without the .git part *)
+    check "git@github.com:username/repo" (Some ("username", "repo"));
+    check "git@github.com:user-name/repo" (Some ("user-name", "repo"));
+    check "git@github.com:user-name-123/repo" (Some ("user-name-123", "repo"));
+    check "git@github.com:123/repo" (Some ("123", "repo"));
+    check "wrong" None;
+    check "https://github.com/username/repo.git" (Some ("username", "repo"));
+  ]
+
+let suite = ("uri", test_v)
```
NathanReb: This should be "Github_uri" and the file should be named `test_github_uri.ml` I think!

gpetiot: done

gpetiot:
I improved the code a bit. So far `split` is still used for `github.io` uris. I do'nt know if we should integrate it in this abstraction, but I'm afraid we will end up with something similar to `Uri.t` and I think it's good to distinguish repo uris and doc uris.

NathanReb:
@gpetiot I thought we merged this somehow, does it still need a review? Sorry for the delay, I'm a bit confused as to why we haven't merged this yet!

gpetiot:
I'm not satisfied with the API yet, it feels half-done to me. I will refresh it and improve it when ocamlformat gives me more air.

NathanReb:
Sure, there's no rush so don't worry about it!

gpetiot:
The last thing I would like to improve is the `Github_uri.split` function, to replace it with other bits of the new `Github_uri` API

tests/bin/no_doc/run.t
```diff
@@ -195,14 +195,15 @@ We do the whole process, calling publish doc implicitely should succeed
...
+    [-] Pushing tag 0.1.0 to git@github.com/foo/whatever.git
+    -: exec: git --git-dir .git push --force git@github.com/foo/whatever.git 0.1.0
     ...
-    [?] Create release 0.1.0 on git@github.com:foo/whatever.git? [Y/n]
-    [-] Creating release 0.1.0 on git@github.com:foo/whatever.git via github's API
+    [?] Create release 0.1.0 on git@github.com/foo/whatever.git? [Y/n]
+    [-] Creating release 0.1.0 on git@github.com/foo/whatever.git via github's API
     ...
-    -: exec: curl -u foo:${token} -L -s -S -K - -D - --data
+    -: exec: curl -u foo/whatever/releases/download/0.1.0:${token} -L -s -S -K - -D
```
gpetiot, line 205: investigating this

gpetiot:
I don't think it's ready to be merged, more tests are probably needed, but it's ready for another round of review if you have the time @NathanReb @pitag-ha

lib/pkg.ml
```diff
@@ -312,61 +285,33 @@ let distrib_file ~dry_run p =
...
-            >>| (fun p -> (user, repo, Fpath.rem_empty_seg p))
-            |> R.reword_error_msg (fun _ -> uri_error uri) )
+module Github = struct
+  let distrib_uri p =
+    distrib_uri p >>= fun uri ->
+    match Github_uri.Distrib.of_string uri with
+    | Ok uri -> Ok uri
+    | Error _ ->
+        R.error_msgf
+          "Could not derive user and repo from opam dev-repo field value %a; \
```
NathanReb, line 300: The message is a bit misleading, how about something like:
```
One of homepage or dev-repo in xx.opam was expected to point to a github repository
```

lib/pkg.ml
```diff
@@ -312,61 +285,33 @@ let distrib_file ~dry_run p =
...
+           expected the pattern $SCHEME://$HOST/$USER/$REPO[.$EXT][/$DIR]"
+          String.dump uri
+
+  let doc_uri p =
+    doc_uri p >>= fun uri ->
+    match Github_uri.Doc.of_string uri with
+    | Ok uri -> Ok uri
+    | Error _ ->
+        R.error_msgf
+          "Could not derive publication directory $PATH from opam doc field \
```
NathanReb, line 310: Same here, how about something like:
```
The doc field  in xx.opam  was expected to point to a github.io URL
```

tests/bin/no_doc/run.t
```diff
@@ -195,12 +195,12 @@ We do the whole process, calling publish doc implicitely should succeed
     [-] Publishing to github
     -: exec: git --git-dir .git rev-parse --verify 0.1.0
     -: exec: git --git-dir .git rev-parse --verify 0.1.0
-    [?] Push tag 0.1.0 to git@github.com:foo/whatever.git? [Y/n]
-    [-] Pushing tag 0.1.0 to git@github.com:foo/whatever.git
-    -: exec: git --git-dir .git push --force git@github.com:foo/whatever.git 0.1.0
+    [?] Push tag 0.1.0 to git@github.com/foo/whatever.git? [Y/n]
+    [-] Pushing tag 0.1.0 to git@github.com/foo/whatever.git
+    -: exec: git --git-dir .git push --force git@github.com/foo/whatever.git 0.1.0
```
NathanReb, line 200: this should still be a `:` and not a `/`

tests/bin/no_doc/run.t
```diff
@@ -195,12 +195,12 @@ We do the whole process, calling publish doc implicitely should succeed
...
     ...
-    [?] Create release 0.1.0 on git@github.com:foo/whatever.git? [Y/n]
-    [-] Creating release 0.1.0 on git@github.com:foo/whatever.git via github's API
+    [?] Create release 0.1.0 on git@github.com/foo/whatever.git? [Y/n]
```
NathanReb, line 202: Same here!

lib/pkg.ml
```diff
@@ -8,25 +8,18 @@ open Bos_setup
 
 (* Misc *)
 
-let uri_domain uri =
-  match Text.split_uri uri with
-  | None -> []
-  | Some (_, host, _) -> List.rev (String.cuts ~sep:"." host)
-
 let uri_sld uri =
```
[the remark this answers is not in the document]
gpetiot: it looks like it's second-level domain, it still fits the code

lib/pkg.ml

lib/pkg.ml

lib/pkg.ml
```diff
@@ -207,30 +194,22 @@ let licenses p =
 let dev_repo p =
   opam_field_hd p "dev-repo" >>= function
   | None -> Ok None
-  | Some r -> (
-      let uri = chop_git_prefix r in
-      match String.cut ~sep:"https://github.com/" uri with
-      | Some ("", path) -> Ok (Some ("git@github.com:" ^ path))
-      | _ -> Ok (Some uri) )
```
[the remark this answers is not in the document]
gpetiot: This is now fixed, the linting of non-github URIs is now FAIL but `dune-release distrib` doesn't fail and the error is only raised with `dune-release publish`

lib/lint.ml
```diff
@@ -102,11 +102,24 @@ let lint_opam_doc pkg =
...
 
 let lint_opam_home_and_dev pkg =
-  lint_res
-    ~msgf:(fun l ->
-      l "opam fields %a and %a can be parsed by dune-release" pp_field
-        "homepage" pp_field "dev-repo")
-    (Pkg.distrib_user_and_repo pkg)
+  ( match Pkg.distrib_uri pkg with
+  | Error _ ->
+      report_status `Fail (fun l ->
```
NathanReb, line 107: Since we're not going to exit 0 we should make it a warning I guess, what do you think?

We really need to think about the lint part and decide what should make something fatal or not. I'm thinking that maybe we should remove the lint step from the `distrib` command and run separately or as part of bistro. That way the `--skip-lint` would only be a flag for bistro and people unhappy with dune-release's linting can skip it if they know what they're doing.

lib/pkg.ml
```diff
@@ -246,21 +225,24 @@ let distrib_uri_of_dev_repo p =
...
+      path_of_distrib p >>= fun path' ->
+      Fpath.of_string path' >>= fun path' ->
+      Github_uri.Homepage.of_string s >>| fun { user; repo; scheme; path } ->
+      match path with
+      | Some path ->
+          let path = Fpath.append path path' in
+          Some (Github_uri.Distrib.make ~user ~repo ~scheme ~path)
+      | None -> Some (Github_uri.Distrib.make ~user ~repo ~scheme ~path:path') )
 
 let distrib_uri ?(raw = false) p =
```
NathanReb, line 247: Is there any reason why this is still exposed and used elswhere in the code considering there has to be a github uri for this function to return `Ok _` ?

lib/lint.ml
```diff
@@ -102,11 +102,24 @@ let lint_opam_doc pkg =
...
+          l
+            "Fields homepage and dev-repo are not valid github URIs, \
+             dune-release will not be able to automatically publish the \
+             package.")
+  | Ok "" ->
+      report_status `Fail (fun l ->
+          l "Skipping homepage and dev-repo fields linting, no field found")
+  | Ok _ ->
+      let pass = R.is_ok (Pkg.Github.distrib_uri pkg) in
```
NathanReb, line 116: Is this here in case the `--dist-uri` option was passed or is there another reason?

lib/pkg.ml
```diff
@@ -246,21 +225,24 @@ let distrib_uri_of_dev_repo p =
...
 
 let distrib_uri_of_homepage p =
-  opam_homepage_sld p >>= function
+  opam_homepage p >>= function
   | None -> Ok None
-  | Some (uri, _) ->
-      path_of_distrib p >>| fun path -> Some (uri_append uri path)
+  | Some s -> (
+      path_of_distrib p >>= fun path' ->
+      Fpath.of_string path' >>= fun path' ->
```
pitag-ha, line 239:
```suggestion
      path_of_distrib p >>= Fpath.of_string >>= fun path' ->
```

Is that the same?

lib/pkg.ml
```diff
@@ -207,30 +194,22 @@ let licenses p =
...
   opam_field_hd p "dev-repo" >>= function
   | None -> Ok None
-  | Some r -> (
-      let uri = chop_git_prefix r in
-      match String.cut ~sep:"https://github.com/" uri with
-      | Some ("", path) -> Ok (Some ("git@github.com:" ^ path))
-      | _ -> Ok (Some uri) )
+  | Some r ->
+      Github_uri.Repo.of_string r >>| fun uri ->
+      Some (Github_uri.Repo.to_string { uri with scheme = GIT })
```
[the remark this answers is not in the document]
gpetiot: This is maintained in this PR to not change the behavior too much but I agree we would need to preserve the initial uri scheme in another PR.

lib/pkg.ml
```diff
@@ -246,21 +225,24 @@ let distrib_uri_of_dev_repo p =
   opam_field_hd p "dev-repo" >>= function
   | None -> Ok None
   | Some dev_repo ->
-      let dev_repo =
-        match String.cut ~sep:"git@github.com:" dev_repo with
-        | Some ("", path) -> "https://github.com/" ^ chop_ext path
-        | _ -> (
-            match String.cut ~sep:"git+ssh://git@github.com/" dev_repo with
-            | Some ("", path) -> "https://github.com/" ^ chop_ext path
-            | _ -> chop_git_prefix (chop_ext dev_repo) )
```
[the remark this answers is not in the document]
gpetiot: checked in the non-github uri cram tests

lib/pkg.ml
gpetiot: fixed

lib/lint.ml
gpetiot: This just indicates whether we might face an issue later on if the distrib uri is not a github uri, but without failing.

gpetiot:
Closing this one, it's too far from master and the new check command includes a lot of changes from this PR

PR #217, opened 2020-03-03
Split unit tests of tag functions

gpetiot:


gpetiot:
I agree the level of these tests is not ideal, I can work on this in another PR.

PR #220, opened 2020-04-06
Add a config subcommand to display and edit global configuration

### LINKED ISSUE #185: Add a `dune-release config` subcommand
Exposing commands to show and edit the global `dune-release` configuration would make it easier for user to tweak should the need to do so arise and would also allow us to change the format, the filename etc... as we have a clean external API to work with it.

I suggest the following spec:
- `dune-release config` or `dune-release config show` displays the global configuration in a human readable format
- `dune-release config set <key> <value>` sets `<key>` to `<value>` in the configuration. The man page for `dune-release config` should document the various keys and what they correspond to.

NathanReb:
Fixes #185 

This PR add a new subcommand to display and edit the global dune-release configuration, along with a bit of documentation as to what each of the fields is used for.

This should make the global config and its components more discoverable for users.

While working on this I noticed that the user field might not be of any use and that it may even cause bugs if it differs from the one in the remote opam-repo URL. We should take a look at this separately!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,13 @@
 
 ### Added
 
+<<<<<<< HEAD
 - Add command `delegate-info` to print information needed by external 
   release scripts (#221, @pitag-ha)
+=======
+- Add a `dune-release config` subcommand to display and edit the global
+  configuration (#220, @NathanReb).
+>>>>>>> c484560... Add a config subcommand to display and edit global configuration
 
 ### Changed
 
--- a/bin/config.ml
+++ b/bin/config.ml
@@ -1,0 +1,144 @@
+open Dune_release
+
+let invalid_config_key key =
+  Rresult.R.error_msgf "%S is not a valid global config field" key
+
+let show_val = function None -> "<unset>" | Some x -> x
+
+let log_val string_opt =
+  Logs.app (fun l -> l "%s" (show_val string_opt));
+  Ok ()
+
+let show key =
+  let open Rresult.R.Infix in
+  Config.load () >>= fun config ->
+  match key with
+  | None ->
+      let pretty_fields = Config.pretty_fields config in
+      StdLabels.List.iter pretty_fields ~f:(fun (key, value) ->
+          Logs.app (fun l -> l "%s: %s" key (show_val value)));
+      Ok ()
+  | Some "user" -> log_val config.user
+  | Some "remote" -> log_val config.remote
+  | Some "local" -> log_val (Stdext.Option.map ~f:Fpath.to_string config.local)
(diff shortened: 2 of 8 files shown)


lib/stdext.ml
```diff
@@ -47,3 +47,21 @@ module Unix = struct
...
+module List = struct
+  let filter_map ~f l =
+    let rec fmap acc = function
+      | [] -> List.rev acc
+      | hd :: tl -> (
+          match f hd with None -> fmap acc tl | Some x -> fmap (x :: acc) tl )
+    in
+    fmap [] l
+
+  let filter_none l = filter_map ~f:(fun x -> x) l
```
gpetiot, line 66: It doesn't seem to be used.

lib/stdext.ml
```diff
@@ -47,3 +47,21 @@ module Unix = struct
   let read_line ?(echo_input = true) () =
     maybe_echo_input echo_input read_line ()
 end
+
+module Option = struct
+  let map ~f = function None -> None | Some x -> Some (f x)
+
+  let is_some = function None -> false | Some _ -> true
```
gpetiot, line 54: It doesn't seem to be used.

bin/config.ml
```diff
@@ -0,0 +1,144 @@
...
+    "The action to perform, either $(b,show) the config or $(b,set) a config \
+     field"
+  in
+  Cmdliner.Arg.(value & pos 0 string "show" & info ~doc ~docv [])
+
+let key =
+  let docv = "KEY" in
+  let doc =
+    "The configuration field to set or print. For $(b,show), if no key is \
+     provided, the entire config will be pretty printed."
```
gpetiot, line 133: Just "printed" I think.

CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 ### Added
 
+- Add a `dune-release config` subcommand to display and edit the global
```
gpetiot, line 5: Wouldn't it be better to call it "local configuration" instead of "global configuration" ?

bin/config.ml
NathanReb: Yep, good call!

CHANGES.md
NathanReb: Yeah I see why it can be confusing. By global I meant system wide or user config, as opposed to the per-project config we might add soon. Maybe we should use a different terminology here to avoid confusion.

What do you think?

gpetiot: Oh right, I guess it makes sense then.

NathanReb:
I'll try to add e2e tests in a subsequent PR!

Issue #222, opened 2020-04-16
Add x-commit-hash field to the opam file when releasing


NathanReb:
We should incorporate the commit hash in the metadata of the package when releasing, in addition to make things easier to reproduce (tags aren't always handled as seriously as they should) it will help other git based tools to pull in the right source tree, in particular it will be very useful for duniverse!

gpetiot:
Closed by #224

PR #224, opened 2020-04-20
Add x-commit-hash field to the opam file when releasing

gpetiot:
First attempt at fixing #222


tests/test_opam_file.ml
```diff
@@ -27,7 +31,8 @@ description: "This package is nice"
 |}
   in
   let expected_v1 =
-    {|opam-version: "2.0"
+    {|x-commit-hash: "6814f8b26946358c72b926706f210025f36619b0"
```
NathanReb, line 34: I'm surprised it's printed as the first field!

gpetiot: I looked this up, it does'nt look like we have much control upon this, we are just calling write_to_string

NathanReb: Yeah I wasn't expecting we'd have control over this. It doesn't matter much anyway!

NathanReb:
@gpetiot any news on this front? Is it still a draft?

gpetiot:
I haven't look at this in a while, I think I remember the e2e test was incomplete (the opam file tested is not the right one), plus testing with different tag settings is not done either.

tests/bin/x-commit-hash/run.t
```diff
@@ -0,0 +1,99 @@
...
+
+We create an opam package:
+
+    $ dune-release opam pkg --dry-run
+    [-] Creating opam package description for whatever
+    dune-release: [WARNING] Could not find _build/whatever-0.1.0.url.
+    dune-release: [WARNING] using https://github.com/foo/whatever/releases/download/0.1.0/whatever-0.1.0.tbz for as url.src. Note that it might differ from the one generated by Github
+    [+] Wrote opam package description _build/whatever.0.1.0/opam
+    $ cat _build/whatever.0.1.0/opam
+    cat: _build/whatever.0.1.0/opam: No such file or directory
```
gpetiot, line 98: That's weird, this file should exist and contain the x-commit-hash field

NathanReb: Hmm yes indeed, any idea what's wrong here? Could you look into it?

NathanReb: Maybe try to remove the `--dry-run` argument!

gpetiot: good catch

gpetiot: although this field contains the commit id and not the tag if there is one, is this what you had in mind for #222 ?

NathanReb: That's what I had in mind yes!

kit-ty-kate:
I wasn't aware of this but it seems rather useful. Thanks!

PR #226, opened 2020-05-04
Command `tag`: improve error and status messages

### LINKED ISSUE #225: tag subcommand fails with unhelpful error message when the tag exists
Running `dune-release tag` fails if the tag already exists:
```
$ dune-release tag
[-] Extracting tag from first entry in CHANGES.md
[-] Using tag "v0.2.0+test"
[?] Create git tag v0.2.0+test for HEAD? [Y/n]

dune-release: [ERROR] The following command exited with code 128:
                      git --git-dir .git tag -a -m Distribution v0.2.0+test
                      v0.2.0+test HEAD
```

This isn't very helpful. I think that to the very least we should:
1. clean up the error logs in those cases, I'm really not so fond of `Logs` default error reporting.
2. Show external commands stderr/stdout when they fail

In this particular case we could go even further and check whether the tag exists before tagging.If it exists and points to the current `HEAD`, we could successfully exit, informing the user that the tag already exists and points to the right place. If it points to somewhere else, we could fail and inform the user why we did.

pitag-ha:
Fixes #225.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
 
 ### Changed
 
+- Command `tag`: improve error and log messages by comparing the provided 
+  commit with the commit correspondent to the provided tag (#226, @pitag-ha)
+
 ### Deprecated
 
 - Deprecate opam 1.x (#195, @gpetiot)
--- a/bin/tag.ml
+++ b/bin/tag.ml
@@ -8,20 +8,52 @@ open Bos_setup
 open Dune_release
 
 let vcs_tag tag ~dry_run ~commit_ish ~force ~sign ~delete ~msg ~yes =
-  let msg = match msg with None -> strf "Distribution %s" tag | Some m -> m in
   Vcs.get () >>= fun repo ->
-  match delete with
-  | true ->
-      Prompt.confirm_or_abort ~yes ~question:(fun l ->
-          l "Delete tag %a?" Text.Pp.version tag)
-      >>= fun () ->
+  Vcs.commit_id ~dirty:false ~commit_ish repo
+  |> R.reword_error (fun (`Msg msg) ->
+         R.msgf "Invalid commit-ish %s: %s" commit_ish msg)
+  >>= fun commit ->
+  let tag_commit_opt = Vcs.tag_points_to ~tag repo in
+  match (tag_commit_opt, delete) with
+  | Some tag_commit, true ->
+      let question =
+        if tag_commit = commit then
+          Prompt.confirm_or_abort ~yes ~question:(fun l ->
+              l "Delete tag %a?" Text.Pp.version tag)
+        else
+          Prompt.confirm_or_abort_neg ~yes ~question:(fun l ->
+              l
+                "%a Tag %a does not point to the commit you've provided \
+                 (default: HEAD). Do you want to delete it anyways?"
+                Fmt.(styled `Red string)
+                "Warning:" Text.Pp.version tag)
+      in
+      question >>= fun () ->
       Vcs.delete_tag ~dry_run repo tag >>| fun () ->
       App_log.success (fun m -> m "Deleted tag %a" Text.Pp.version tag)
-  | false ->
+  | None, true ->
+      Ok
+        (App_log.status (fun apply_log_l ->
+             apply_log_l "Nothing to be deleted: there is no tag %a."
+               Text.Pp.version tag))
+  | Some tag_commit, false ->
+      if tag_commit = commit then
+        Ok
+          (App_log.status (fun apply_log_l ->
+               apply_log_l "Nothing to be done: tag already exists."))
+      else
+        R.error_msgf
+          "A tag with name %a already exists, but points to a different \
+           commit. You can delete that tag using the `-d` flag."
+          Text.Pp.version tag
+  | None, false ->
       Prompt.confirm_or_abort ~yes ~question:(fun l ->
           l "Create git tag %a for %a?" Text.Pp.version tag Text.Pp.commit
             commit_ish)
       >>= fun () ->
+      let msg =
+        match msg with None -> strf "Distribution %s" tag | Some m -> m
+      in
       Vcs.tag repo ~dry_run ~force ~sign ~msg ~commit_ish tag >>| fun () ->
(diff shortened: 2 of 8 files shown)


lib/vcs.ml
```diff
@@ -400,6 +400,14 @@ let tag_exists ~dry_run r tag =
   | (`Git, _, _) as r -> git_tag_exists r ~dry_run tag
   | `Hg, _, _ -> failwith "TODO"
 
+let tag_points_to_commit r tag commit_ish =
```
NathanReb, line 403: I'm not sure this function is required. What would you think about extracting a function:
```ocaml
(** [tag_points_to t tag] returns the commit hash [tag] points to or [None] if it doesn't exist *)
tag_points_to : dry_run: bool -> t -> string -> string option
```
and implement `tag_exists` on top of it. We can then clean the tag  command code a bit so that it starts by getting the `commit_ish` actual commit. Note that this we should fail if it doesn't exist anyway so might as well do it early. It would then get the commit for the tag and behave accordingly.

bin/tag.ml
```diff
@@ -10,22 +10,54 @@ open Dune_release
...
   Vcs.get () >>= fun repo ->
+  let tag_exists = Vcs.tag_exists ~dry_run:false repo tag in
   match delete with
   | true ->
-      Prompt.confirm_or_abort ~yes ~question:(fun l ->
-          l "Delete tag %a?" Text.Pp.version tag)
-      >>= fun () ->
-      Vcs.delete_tag ~dry_run repo tag >>| fun () ->
-      App_log.success (fun m -> m "Deleted tag %a" Text.Pp.version tag)
+      if tag_exists then
```
NathanReb, line 16: Following my suggestion regarding `tag_points_to`, I think this whole function could be simplified a bit, maybe as a single pattern matching over `delete`, `commit_ish` and the optional commit the tag points to.

It is a bit convoluted at the moment and makes it hard to follow. What do you think?

bin/tag.ml
```diff
@@ -10,22 +10,54 @@ open Dune_release
...
+                   (default: HEAD). Do you want to delete it anyways?"
+                  Fmt.(styled `Red string)
+                  "Warning:" Text.Pp.version tag)
+        in
+        question >>= fun () ->
+        Vcs.delete_tag ~dry_run repo tag >>| fun () ->
+        App_log.success (fun m -> m "Deleted tag %a" Text.Pp.version tag)
+      else
+        Ok
+          (App_log.success (fun apply_log_l ->
```
NathanReb, line 36: I'm wondering whether this should be a success or status since nothing actually happened. What do you think?

lib/prompt.ml
```diff
@@ -1,22 +1,33 @@
...
   App_log.question (fun l ->
-      f (fun ?header ?tags fmt -> l ?header ?tags (fmt ^^ " [Y/n]")))
+      f (fun ?header ?tags fmt ->
+          l ?header ?tags (fmt ^^ Scanf.format_from_string options "")))
 
-let confirm ~question ~yes =
-  let rec loop () =
-    ask question;
+let rec loop ~question ~default_answer ~yes =
+  ask question ~default_answer;
```
NathanReb, line 10: As mentioned below, we shouldn't "ask" if `--yes` was passed and just go straight to the point so this should be moved to the else branch I think.

bin/tag.ml
pitag-ha: Yes, status definitely seems more appropriate here. I've also replaced `success` by `status` in the case of trying to create a tag that already exists providing a "right" commit-ish.

lib/vcs.ml
pitag-ha: Yes, that structure you're suggesting is far nicer! I'm just not sure if I understand what you have in mind for `tag_points_to` and `tag_exists`. Could you have another look (once I push the changes)?

lib/prompt.ml
```diff
@@ -1,22 +1,33 @@
-let ask f =
+type default = Yes | No
+
+let ask f ~default_answer =
+  let options = match default_answer with Yes -> " [Y/n]" | No -> " [y/N]" in
   App_log.question (fun l ->
-      f (fun ?header ?tags fmt -> l ?header ?tags (fmt ^^ " [Y/n]")))
+      f (fun ?header ?tags fmt ->
+          l ?header ?tags (fmt ^^ Scanf.format_from_string options "")))
```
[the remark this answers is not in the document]
pitag-ha: Cool! That works.

CHANGES.md
```diff
@@ -7,9 +7,16 @@
 - Add command `delegate-info` to print information needed by external
   release scripts (#221, @pitag-ha)
 - Use Curly instead of Cmd to interact with github (#202, @gpetiot)
+- Add `Prompt.confirm_or_abort_neg`: Like `[Prompt.confirm_or_abort]`,
```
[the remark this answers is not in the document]
pitag-ha: That's good to know. Thanks!

bin/tag.ml
pitag-ha: Yes, that actually makes it far nicer! :) Thanks.

tests/bin/delegate_info/run.t
```diff
@@ -23,6 +23,7 @@ We need to set up a git project for dune-release to work properly
     $ dune-release tag -y
     [-] Extracting tag from first entry in CHANGES.md
     [-] Using tag "0.1.0"
+    [?] Create git tag 0.1.0 for HEAD? [Y/n]
```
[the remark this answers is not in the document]
pitag-ha: Ok, the old behaviour is back :)

lib/prompt.ml
pitag-ha: Yeah, asking a question when auto-confirming does feel a bit off... The reason why I did that was to be able to E2E test the different behaviour of `tag -d` when providing a commit-ish, that corresponds to the provided tag, or providing a "wrong" commit-ish (the difference lies in the question that asks for confirmation and in the default answer to that question). Do you think I should test that part of `tag` by adding a (general) flag `--answer-default` or should I simply not test that part?

PR #227, opened 2020-05-05
Removing dead code

gpetiot:
(Spring cleaning)

Less code to browse, less `grep` results, easier to maintain.

### DIFF
--- a/lib/archive.mli
+++ b/lib/archive.mli
@@ -35,9 +35,6 @@ val bzip2 :
   dry_run:bool -> ?force:bool -> dst:Fpath.t -> string -> (unit, R.msg) result
 (** [bzip2 dst s] compresses [s] to [dst] using bzip2. *)
 
-val ensure_tar : unit -> (unit, R.msg) result
-(** [ensure_tar ()] makes sure the [tar] utility is available. *)
-
 val untbz : dry_run:bool -> ?clean:bool -> Fpath.t -> (Fpath.t, R.msg) result
 (** [untbz ~clean ar] untars the tar bzip2 archive [ar] in the same directory as
     [ar] and returns a base directory for [ar]. If [clean] is [true] (defaults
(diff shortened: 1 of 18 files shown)


lib/archive.mli
```diff
@@ -35,9 +35,6 @@ val bzip2 :
   dry_run:bool -> ?force:bool -> dst:Fpath.t -> string -> (unit, R.msg) result
 (** [bzip2 dst s] compresses [s] to [dst] using bzip2. *)
 
-val ensure_tar : unit -> (unit, R.msg) result
```
NathanReb, line 38: Hmm I'd be curious to know why this is not used anymore. As long as we have a decent error message if `tar`'s missing I'm fine with it!

lib/pkg.ml
```diff
@@ -518,7 +518,7 @@ let distrib_version_opam_files ~dry_run ~version =
   infer_pkg_names Fpath.(v ".") [] >>= fun names ->
   List.fold_left
     (fun acc name ->
-      acc >>= fun _acc ->
+      acc >>= fun () ->
```
NathanReb, line 521: Maybe we should extract a `Stdext.Result.List.iter` for this. I'm fine with leaving it as it is but this might be a bit confusing. Hopefully we'll soon be able to use `stdune`!

lib/xdg.mli
```diff
@@ -6,10 +6,4 @@
 val config_dir : string
 (** The directory where the application should read/write config files. *)
 
-val data_dir : string
```
NathanReb, line 9: I think we could leave `Xdg` out of this since it's kind of vendored. What do you think?

NathanReb: Maybe the right solution here actually is to make it a standalone library in a `vendors/` dir at the root of the project!

lib/archive.mli
gpetiot: `tar` is only used in `Archive.untbz` that already checks that tar is present itself

lib/xdg.mli
gpetiot: I will revert it and we can do that in another PR.

lib/pkg.ml

Issue #228, opened 2020-05-07
Clean up error logs


pitag-ha:
At the moment, error logs look as follows:
```
dune-release: [ERROR] The following command exited with code 128:
                      git --git-dir .git tag -a -m Distribution v0.2.0+test
                      v0.2.0+test HEAD
```
We should
- get rid of "dune-release:" at the beginning: when a user runs a `dune-release`-command which fails, starting the error log with "dune-release" doesn't contribute any useful information;
- not indent any line.

That way, the command that's failed might not get split into two lines.

pitag-ha:
I've split up #225 into three parts. This issue is one of them.

PR #231, opened 2020-05-11
Improve logs when external command fails

### LINKED ISSUE #229: Include external command error logs into `dune-release` error logs
When `dune-release` executes an external command (git or hg) that fails, the message on the external command's process stderr should also appear in the `dune-release` error message.

Example: now, a failed external command gets reported as follows:
```
The following command exited with code 128: 
git --git-dir .git tag -a -m Distribution v0.2.0+test v0.2.0+test HEAD
```
We should include why that command has failed:
```
The following command exited with code 128 and error message "fatal: tag 'v0.2.0+test' already exists": 
git --git-dir .git tag -a -m Distribution v0.2.0+test v0.2.0+test HEAD 
```

pitag-ha:
Closes #229. Two comments on this PR:
 - The reason why I'm passing the error message from the external command through a (temporary) file is that that's the only way provided by the `bos`-library to capture an error message.
 - I've left error messages coming from failed `mercurial`-commands the way they were for now, because they don't seem fully supported in general. I can improve that if it would make sense.


lib/prompt.mli
```diff
@@ -11,3 +11,7 @@ val confirm_or_abort :
   question:('a, unit) Logs.msgf -> yes:bool -> (unit, Rresult.R.msg) result
 (** Same as [confirm] but returns [Ok ()] for yes and
     [Error (`Msg "Aborting on user demand")] for no *)
+
+val confirm_or_abort_neg :
+  question:('a, unit) Logs.msgf -> yes:bool -> (unit, Rresult.R.msg) result
+(** Like [confirm_or_abort], but defaults to no when hitting enter *)
```
gpetiot, line 17: What do you think of exporting the type `type choice = Yes | No` so that `?default_answer` can be a parameter of `confirm`/`confirm_or_abort` instead of adding a new function?

pitag-ha: Yes, I that's a good idea. I was trying to mess as little as possible with already written functions, but I think you're right.

As a site-note: This is quite a mess. Sorry! That commit you're referring to was part of my last PR  (and shouldn't be part of this one): https://github.com/ocamllabs/dune-release/pull/226/commits/149578a3c2da42bdfe7cf06acaa809c64eaac78f
I'll remove that commit from this PR and implement your suggestion in a separate one, if that makes sense.

pitag-ha:
One doubt: I think it might improve readability and coherence if `Sos.run_out` and `Sos.run_out_err` had a similar return type. Should I adapt the one of `Sos.run_out` to the one I've chosen for `Sos.run_out_err` (I, personally, find that one more readable)? Or the other way around?

lib/vcs.ml
```diff
@@ -64,28 +64,49 @@ let find_git () =
...
-  OS.Cmd.(run_out ~err:err_null status |> out_string) >>= function
-  | "", (_, `Exited 0) -> Ok false
-  | _, (_, `Exited 0) -> Ok true
-  | _, (_, st) -> cmd_error status st
+  Sos.run_out_err ~dry_run:false status ~default:D.string OS.Cmd.out_string
+  >>= fun response ->
+  match (response.output, response.status) with
+  | "", `Exited 0 -> Ok false
+  | _, `Exited 0 -> Ok true
+  | _, _ -> cmd_error status ~err_msg:response.err_msg response.status
```
NathanReb, line 109: This is just a style note but you could have written it like this:
```suggestion
  Sos.run_out_err ~dry_run:false status ~default:D.string OS.Cmd.out_string
  >>= function
  | {output = ""; status = `Exited 0; _} -> Ok false
  | {status = `Exited 0; _} -> Ok true
  | {err_msg; status; _} -> cmd_error status ~err_msg status
```

lib/vcs.ml
```diff
@@ -64,28 +64,49 @@ let find_git () =
...
-  | `Exited c ->
-      R.error_msgf "The following command exited with code %d:@\n%a" c Cmd.pp
-        cmd
-  | `Signaled c ->
-      R.error_msgf "The following command exited with signal %d:@\n%a" c Cmd.pp
-        cmd
+let cmd_error cmd ?err_msg =
+  match err_msg with
+  | None | Some "" -> (
+      function
```
NathanReb, line 70: I think this doesn't read very well. What would you think about explicitly adding the extra `status` argument above and wrap this in a single pattern-matching?

lib/vcs.ml
```diff
@@ -64,28 +64,49 @@ let find_git () =
...
       | _ -> Ok None )
 
-let cmd_error cmd = function
-  | `Exited c ->
-      R.error_msgf "The following command exited with code %d:@\n%a" c Cmd.pp
-        cmd
-  | `Signaled c ->
-      R.error_msgf "The following command exited with signal %d:@\n%a" c Cmd.pp
-        cmd
+let cmd_error cmd ?err_msg =
```
NathanReb, line 67: What would you think about moving this to `Sos` instead? I think it could benefit to more places in the code than just `Vcs` and it seems generic enough!

Also one thing I'm wondering is why is `err_msg` optional? Shouldn't it be mandatory? I feel like if one wants to report a command failure they should have run it with `run_out_err` and should always pass this function the command's stderr.

lib/vcs.ml
pitag-ha: Oh, nice. I remember I read about record pattern matching, thought it was cool and then forgot it again. Thanks!

lib/vcs.ml
pitag-ha: Yes, that indeed looks better :)

lib/vcs.ml
pitag-ha: Ok, I've moved it to `Sos` . 

The reason why `err_msg` was optional was because there are still commands for which I've consciously decided not to print out the error message from the external command (`git_file_is_dirty` and `run_hg`). As you've suggested on Slack, I've now made the parameter `err_msg` mandatory (but being an option type), so that we can't just forget to pass the error message to the function. I think that's a good idea.

lib/prompt.mli
pitag-ha: See #233

lib/sos.ml
```diff
@@ -61,6 +61,20 @@ let show ?sandbox ?(action = `Skip) fmt =
       Ok ())
     fmt
 
+let cmd_error cmd err_msg status =
+  match (err_msg, status) with
+  | None, `Exited c | Some "", `Exited c ->
```
NathanReb, line 66: You can simplify this into:
```suggestion
  | (None | Some ""), `Exited c ->
```

that of course apply to the other case below!

lib/vcs.ml
```diff
@@ -64,28 +64,29 @@ let find_git () =
...
       | _ -> Ok None )
 
-let cmd_error cmd = function
-  | `Exited c ->
-      R.error_msgf "The following command exited with code %d:@\n%a" c Cmd.pp
-        cmd
-  | `Signaled c ->
-      R.error_msgf "The following command exited with signal %d:@\n%a" c Cmd.pp
-        cmd
+module D = struct
```
NathanReb, line 67: How about naming this module `Default` so it's unambiguous? Sorry I wanted to suggest that during the first review round but then forgot about it!

lib/sos.ml
pitag-ha: Cool, that's a very nice pattern match. I like it :)

tests/bin/external_errors/run.t
```diff
@@ -0,0 +1,38 @@
...
+
+Let's make `dune-release` run a `git`-command that's doomed to fail. After the customized error line, the error log should contain 
+- the exit code/signal,
+- the external command that has failed,
+- the error message the external command has posted on its stderr.
+
+    $ dune-release tag --commit=1
+    [-] Extracting tag from first entry in CHANGES.md
+    [-] Using tag "0.1.0"
+    dune-release: [ERROR] Due to invalid commit-ish `1`:
```
samoht, line 32: Really nice error message! Maybe the formatting could improved a bit by not putting this in a formatting box? e.g. something like:

```
dune-release: [ERROR] 
  Due to invalid commit-ish `1`:
  Exit code 128 from command
    `git --git-dir .git rev-parse --verify 1^{commit}`:
  fatal: Needed a single revision
```

lib/vcs.ml
pitag-ha: Done. It was already called `D` before and I just left the name (and moved it up a bit since I need it in more places now), but I also like calling it `Default` better.

tests/bin/external_errors/run.t
pitag-ha: What do you think about issue  #228 in that context? The idea came from @NathanReb  in issue #225. I wanted to treat that part in a separate PR since it's not directly related to failed external commands.

PR #233, opened 2020-05-15
Fuse confirm_and_abort_neg into confirm_and_abort

pitag-ha:
I'm just restructuring the confirm_and_abort for negative default behavior according to @gpetiot 's suggestion in #231.

### DIFF
--- a/lib/prompt.ml
+++ b/lib/prompt.ml
@@ -1,4 +1,4 @@
-type default = Yes | No
+type answer = Yes | No
 
 let ask f ~default_answer =
   let options : ('a, Format.formatter, unit, unit) format4 =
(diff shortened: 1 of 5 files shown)


lib/prompt.ml
```diff
@@ -21,14 +21,9 @@ let rec loop ~question ~default_answer =
              for the default");
       loop ~question ~default_answer
 
-let confirm ~question ~yes =
-  if yes then true else loop ~question ~default_answer:Yes
+let confirm ~question ~yes ~default_answer =
+  if yes then true else loop ~question ~default_answer
```
gpetiot, line 25: I would make `default_answer` optional and make the default behavior to `Yes` since it's the main usecase (and adding the argument everywhere generates an unnecessary diff)

pitag-ha: Do I have to add the unit as a last parameter for that to work? If so, I think the diff would look similar :/ Or is there some other trick to make that work that I'm not aware of?

gpetiot: Oh yes I didn't pay attention to that

pitag-ha: Would you still do that? Making the `default_answer` optional (defaulting to `Yes`) and adding the unit as positional parameter I mean?

gpetiot: No I guess I would do like you did

lib/prompt.mli
```diff
@@ -1,17 +1,19 @@
-val confirm : question:('a, unit) Logs.msgf -> yes:bool -> bool
-(** Promtps the user for confirmation. [confirm ~question ~yes] uses the message
-    formatting function [question] to format and log a message with the app
-    level and wait for a yes or no answer from the user. Returns [true] for yes.
-    Defaults to yes if the user just press enter. If [yes] then just skip the
-    prompt and returns [true]. E.g.
+type default = Yes | No
```
NathanReb, line 1: Not that it matters much but how about naming this type `answer` instead? I think it makes more sense.

PR #235, opened 2020-06-09
Skip documentation publication when opam's `doc` field is missing 

### LINKED ISSUE #149: dune-release should skip documentation publication when opam's `doc` field is missing

gpetiot:
Fix #149

### DIFF
--- a/bin/publish.ml
+++ b/bin/publish.ml
@@ -33,10 +33,13 @@ let publish_doc ~dry_run ~yes pkg_names pkg =
 
 let pp_field = Fmt.(styled `Bold string)
 
-(* If the `doc` field of the opam file is not set we do not generate nor
-   publish the documentation, except when using a delegate. *)
-let publish_doc ~dry_run ~yes pkg_names pkg =
+(* If `publish doc` is invoked explicitly from the CLI, we should fail if the
+   documentation cannot be published. If it is not called explicitly, we can
+   skip this step if the `doc` field of the opam file is not set, we do not
+   generate nor publish the documentation, except when using a delegate. *)
+let publish_doc ~from_cli ~dry_run ~yes pkg_names pkg =
   match Pkg.doc_uri pkg with
+  | _ when from_cli -> publish_doc ~dry_run ~yes pkg_names pkg
   | Error _ | Ok "" -> (
       match Pkg.delegate pkg with
       | Ok (Some _) -> publish_doc ~dry_run ~yes pkg_names pkg
@@ -63,9 +66,9 @@ let publish_alt ~dry_run pkg kind =
   Pkg.publish_msg pkg >>= fun msg ->
   Delegate.publish_alt ~dry_run pkg ~kind ~msg ~archive
 
-let publish ?build_dir ?opam ?delegate ?change_log ?distrib_uri ?distrib_file
-    ?publish_msg ~name ~pkg_names ~version ~tag ~keep_v ~dry_run
-    ~publish_artefacts ~yes () =
+let publish ?(from_cli = false) ?build_dir ?opam ?delegate ?change_log
+    ?distrib_uri ?distrib_file ?publish_msg ~name ~pkg_names ~version ~tag
+    ~keep_v ~dry_run ~publish_artefacts ~yes () =
   let publish_artefacts =
     match publish_artefacts with [] -> None | v -> Some v
   in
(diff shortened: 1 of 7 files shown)


bin/publish.ml
```diff
@@ -31,6 +30,21 @@ let publish_doc ~dry_run ~yes pkg_names pkg =
...
+   publish the documentation, except when using a delegate. *)
+let publish_doc ~dry_run ~yes pkg_names pkg =
+  App_log.status (fun l -> l "Publishing documentation");
+  match Pkg.doc_uri pkg with
+  | Error _ | Ok "" -> (
+      match Pkg.delegate pkg with
+      | Ok (Some _) -> publish_doc ~dry_run ~yes pkg_names pkg
+      | Error _ | Ok None ->
+          Pkg.name pkg >>= fun name ->
+          App_log.status (fun l -> l "No doc field found for package %s" name);
```
NathanReb, line 43: Regarding the user logs, what would you think about:
- if there's is a doc field, we log `Publishing documentation` and go on with it
- if there's none, we log `Skipping documentation publication as there is no 'doc' field in xxx.opam`

I think it's a bit better than saying that we're going to publish the doc only to say we're not going to.
Might be nice to point to the opam file as well, to avoid confusion about where the doc field is.
I think it seems obvious to us but to new comers or to people that aren't well aware of dune-release's behaviour it might not.

lib/lint.ml
```diff
@@ -88,10 +88,14 @@ let lint_res ~msgf = function
...
 let lint_opam_doc pkg =
-  lint_res
-    ~msgf:(fun l ->
-      l "opam field %a can be parsed by dune-release" pp_field "doc")
-    (Pkg.doc_user_repo_and_path pkg)
+  ( match Pkg.doc_user_repo_and_path pkg with
+  | Ok _ ->
+      report_status `Ok (fun l ->
+          l "opam field %a can be parsed by dune-release" pp_field "doc")
+  | Error _ ->
```
NathanReb, line 95: If we're going to allow missing doc fields, how about not reporting a failure (even though it's non fatal) if it's missing?

What would you think about the following:
- If there's no doc field, we mention it with a `Skipping doc field linting, no doc field for package xxx` or something similar, and succeed
- If there's one, we report whether we can parse it as you did there, in a non-fatal way so that it doesn't hurt people hosting there doc on readthedocs for instance.

bin/publish.ml
NathanReb: Maybe `Skipping documentation publication for package x: no 'doc field in x.opam` is a better message than what I suggested above!

What do you think?

NathanReb:
I'm not too worried about multi-opam repos for now, I think fixing the behaviour for the main use case is already worth something.

What about 2. though?

NathanReb:
As a side note, I think in the future the default behaviour for multi-opam should be: we only publish the doc of packages that have a doc field pointing to ghpages.

Also since we don't necessarily need that doc uri to be able to publish the doc (ie we can infer a good enough one from the dev-repo) might be worth to allow an escape hatch for people that may want to have their doc field pointing elsewhere but still publish their doc to gh-pages as well.

Anyway, I digress but we should open an issue to make sure we don't lose track of these improvement ideas! Could you open one?

gpetiot:
> What about 2. though?

I modified the code to make sure we fail when `publish doc` is invoked explicitly from the CLI and added some tests, let me know what you think

bin/publish.ml
```diff
@@ -43,9 +66,9 @@ let publish_alt ~dry_run pkg kind =
   Pkg.publish_msg pkg >>= fun msg ->
   Delegate.publish_alt ~dry_run pkg ~kind ~msg ~archive
 
-let publish ?build_dir ?opam ?delegate ?change_log ?distrib_uri ?distrib_file
-    ?publish_msg ~name ~pkg_names ~version ~tag ~keep_v ~dry_run
-    ~publish_artefacts ~yes () =
+let publish ?(from_cli = false) ?build_dir ?opam ?delegate ?change_log
```
NathanReb, line 69: This `from_cli` argument seems a bit convoluted to me. I think we could determine whether it was explicitly asked for with the `publish_artifact` argument, ie if it contains `Doc` then it was asked for explicitly. I think this makes things clearer.

I wonder if it wouldn't be worth changing `Pkg.publish_artifact` to `Pkg.default_publish_artifacts` and either use the cli args if they're non empty or the default ones. It would make things much clearer than having to populate the package with the args and always go through `Pkg`. I find it rather confusing.

gpetiot: That's a good catch, I changed this.

PR #239, opened 2020-06-24
Update the README and add more details on each command

### LINKED ISSUE #127: Documentation is fairly incomplete, the tool seems quite hard to use
As seen in https://github.com/garrigue/lablgtk/pull/38#issuecomment-462097726 , migrating packages to use `dune-release` is a very frustrating experience.

Contrary to what is claimed, many critical invariants [such as that `homepage` must be set to github, format of CHANGES, etc...] don't seem to be properly documented, and in the cases that the right hidden magic is not properly set the tool behaves in random ways [such as setting URLs incorreclty, publishing arbitrary changelogs, etc...]

IMHO what works and what does not should be much better documented, likely with a guiding example.

cc: @garrigue

gpetiot:
fix #127 

More details and examples for each command, I also removed the dead links to the API doc.

Here is the preview: https://github.com/gpetiot/dune-release/blob/detail-readme/README.md


README.md
```diff
@@ -26,30 +23,130 @@ dune-release help release # for help about releasing your package
...
+The full documentation of this command is available with
+```
+dune-release help distrib
+```
+
 
-For instance:
+### Publish the distribution and documentation online
 
+Once the distribution archive is created you can now publish it and its documentation online.
```
NathanReb, line 99: Probably worth explaining what publishing means here, ie:
- for the distrib it means: creating a Github release associated with the tag and upload the distrib tarball as a release artifact
- for the doc, publishing the dune generated documentation to gh-pages to be served as a static website on github.io

README.md
```diff
@@ -26,30 +23,130 @@ dune-release help release # for help about releasing your package
...
-       dune-release(1).
+The full documentation of this command is available with
+```
+dune-release help publish
 ```
 
+
+### Create an opam package
+
+The following steps still need the distribution archive created in the preceeding step to be in the build directory. If that's no longer the case but nothing moved in your VCS, you can simply invoke dune-release distrib, it should produce a bit-wise identical archive. If the VCS moved checkout the distribution commit to regenerate the archive or provide, in the subsequent commands, the archive manually via the `--dist-file` option.
```
NathanReb, line 125: I think it might be confusing to explain all this distrib file details. how about explaining that the `pkg` subcommand generates an opam file to be used on opam-repo (ie that includes the DL URI for the distrib tarball and the tarball hash) and the `submit` subcommand creates the PR instead?

README.md
```diff
@@ -26,30 +23,130 @@ dune-release help release # for help about releasing your package
...
+
+This step should be executed prior to running the `dune-release` automated process.
+
+The tagging command of `dune-release` will extract the latest version tag from the package's change log and tag the VCS HEAD commit with it if it is invoked without argument:
+
+```
+dune-release tag
+```
+
+This will only work if the change log follows a certain format. If you do not want to rely on this extraction you can specify it on the command line:
```
NathanReb, line 53: Might be worth adding a section about that to explain what format dune-release expects and what it does with it (ie infer the tag/version and generate a publication message).

README.md
```diff
@@ -26,30 +23,130 @@ dune-release help release # for help about releasing your package
 dune-release help         # for more help
 ```
 
-[doc]: https://samoht.github.com/dune-release/
+Calling `dune-release` without any argument will start the automated release process, composed of the following steps:
+- create the distribution archive;
+- publish it online with its documentation;
+- create an opam package;
+- submit it to OCaml's opam repository.
+
```
NathanReb, line 31: It might be a good place for a short paragraph explaining that if the repo contains multiple opam packages, dune-release will try to release all of them by default (ie one tarball, one gh release but several opam files added to opam-repo).

README.md
```diff
@@ -26,30 +23,130 @@ dune-release help release # for help about releasing your package
...
+The most basic workflow is:
+
+```
+dune-release tag
+dune-release
+```
+
+Each step is refined and explained with more details below.
+If this is your first time using `dune-release` you might choose to run the `dune-release` commands with the `--dry-run` argument, so that no action is actually performed but it allows you to check what would be done. And then run the `dune-release` commands without `--dry-run` when you are sure of what you are doing.
```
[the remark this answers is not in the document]
gpetiot: done

README.md
gpetiot: done

README.md
gpetiot: done

README.md
gpetiot: done

README.md
gpetiot: done

PR #242, opened 2020-06-24
The `git` command used in `publish doc` should check `DUNE_RELEASE_GIT` (even if deprecated) before `PATH` and document runtime dependencies

### LINKED ISSUE #238: Make runtime dependency handling more consistent and document it
I am currently in the midst of packaging `dune-release` for NixOS. Therefore I was looking through the code for runtime
dependencies, i. e. executables that get called. As far as I understand there are two types of runtime dependencies:

1. Some are just expected to be in `PATH`, like coreutils (fair enough) or `ocamlfind`
2. Some have an environment variable configuring the path to their binary, if that is not present, `dune-release` looks for it in `PATH` (git, bzip2, …)

I noticed however that the implementation of the second type of runtime dependencies is in some cases inconsistent: For example, git is configured by `DUNE_RELEASE_GIT` in `lib/vcs.ml`, but assumed to be in `PATH` by `lib/github.ml`.

It would be great if this area of `dune-release` could be made more consistent and documented better. A list of runtime dependencies (or executables called at runtime) and the environment variables that configure them would be great!

gpetiot:
Fix #238

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,3 +168,14 @@ The main differences between `dune-release` and `topkg` are:
 - Bundle everything as a single binary;
 - Use of `Astring`, `Logs`, `Fpath` and`Bos`;
 - Remove the IPC layer (which is used between `topkg` and `topkg-care`);
+
+### Runtime dependencies
+
+- `bzip2` can be set with the `DUNE_RELEASE_BZIP2` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `tar` can be set with the `DUNE_RELEASE_TAR` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `git` can be set with the `DUNE_RELEASE_GIT` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `hg` can be set with the `DUNE_RELEASE_HG` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `opam` can be set with the `HOST_OS_OPAM` variable (deprecated), otherwise it is looked for in `HOST_OS_XBIN/opam` (deprecated), otherwise it is looked for in `opamHOST_OS_SUFF` (deprecated), otherwise it is expected to be in `PATH`;
+- `dune`, `curl`, `cp` and `ocamlfind` are expected to be in `PATH`.
+
+Using these `DUNE_RELEASE_*` and `HOST_OS_*` environment variables to configure the path to these binaries is deprecated since `dune-release.1.4.0`, and will no longer be supported in `dune-release.2.0.0`, it is thus recommended to only rely on the `PATH` variable.
--- a/lib/github.ml
+++ b/lib/github.ml
@@ -119,7 +119,7 @@ let publish_doc ~dry_run ~msg:_ ~docdir ~yes p =
   let create_empty_gh_pages git =
     let msg = "Initial commit by dune-release." in
     let create () =
-      Sos.run_quiet ~dry_run Cmd.(v "git" % "init") >>= fun () ->
+      Sos.run_quiet ~dry_run Cmd.(git % "init") >>= fun () ->
       Vcs.get () >>= fun repo ->
       Ok (git_for_repo repo) >>= fun git ->
       Sos.run_quiet ~dry_run Cmd.(git % "checkout" % "--orphan" % "gh-pages")
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,8 @@
 - Do not write 'version' and 'name' fields in opam file (#200, @gpetiot)
 - Use Yojson to parse github json response and avoid parsing bugs.
   (#177, @gpetiot)
+- The `git` command used in `publish doc` should check `DUNE_RELEASE_GIT` (even
+  if deprecated) before `PATH`. (#242, @gpetiot)
 
 ### Removed
 


README.md
```diff
@@ -63,3 +63,13 @@ The main differences between `dune-release` and `topkg` are:
...
 - Use of `Astring`, `Logs`, `Fpath` and`Bos`;
 - Remove the IPC layer (which is used between `topkg` and `topkg-care`);
+
+### Runtime dependencies
+
+- `bzip2` can be set with the `DUNE_RELEASE_BZIP2` variable (deprecated), otherwise it expected to be in `PATH`;
+- `tar` can be set with the `DUNE_RELEASE_TAR` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `git` can be set with the `DUNE_RELEASE_GIT` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `hg` can be set with the `DUNE_RELEASE_HG` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `curl` is expected to be in `PATH`.
```
sternenseemann, line 73: If `curl` is checked for by `dune-release`, maybe worth it to give `curly` an explicit binary, just to prevent unexpected behavior, maybe overkill though.

gpetiot: That's right, I added this change to the PR.

lib/github.ml
```diff
@@ -173,8 +173,9 @@ let run_with_auth ?(default_body = `Null) ~dry_run ~auth curl_t =
       args
     >>| fun () -> default_body
   else
-    OS.Cmd.must_exist (Cmd.v "curl") >>= fun _ ->
-    match Curly.(run ~args (Request.make ~url ~meth:`POST ())) with
+    OS.Cmd.must_exist (Cmd.v "curl") >>= fun curl ->
+    let exe = Cmd.to_string curl in
```
pitag-ha, line 177: If think `OS.Cmd.must_exist (Cmd.v "curl")` returns `"curl"`. Is that what you want? It seems that the `exe`-argument of `Curly.run` is `curl` anyways. If your intention was to use the explicit binary path (as in `/usr/bin/curl` or so), maybe `OS.Cmd.get_tool ` instead of `OS.Cmd.must_exist` and `Cmd.p curl` instead of `Cmd.to_string curl` would do.

gpetiot: That's a good point. I was thinking that maybe if we passed the string to curly it would not have to guess too much things or check if it exists, but since it uses `"curl"` as default we don't need to do all of this. I'm not sure using the full path to the binary would be useful as it would need to be found in PATH anyway.
I guess it is a matter of whether we want to rely on Curly current default behavior or protect dune-release against changes in the future, and given the activity on Curly I think we are safe relying on current's behavior.

pitag-ha: Oh ok, I think I've misunderstood what you folks meant by "explicit binary" in the conversation about the README.

README.md
gpetiot: After being reviewed it may have been an overkill afterall, so I've reverted this change. Let's keep it this way as long as Curly works with "curl" as default value.

README.md
```diff
@@ -168,3 +168,14 @@ The main differences between `dune-release` and `topkg` are:
 - Bundle everything as a single binary;
 - Use of `Astring`, `Logs`, `Fpath` and`Bos`;
 - Remove the IPC layer (which is used between `topkg` and `topkg-care`);
+
+### Runtime dependencies
+
+- `bzip2` can be set with the `DUNE_RELEASE_BZIP2` variable (deprecated), otherwise it expected to be in `PATH`;
```
pitag-ha, line 174:
```suggestion
- `bzip2` can be set with the `DUNE_RELEASE_BZIP2` variable (deprecated), otherwise it is expected to be in `PATH`;
```

README.md
```diff
@@ -168,3 +168,14 @@ The main differences between `dune-release` and `topkg` are:
...
+- `tar` can be set with the `DUNE_RELEASE_TAR` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `git` can be set with the `DUNE_RELEASE_GIT` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `hg` can be set with the `DUNE_RELEASE_HG` variable (deprecated), otherwise it is expected to be in `PATH`;
+- `opam` can be set with the `HOST_OS_OPAM` variable (deprecated), otherwise it is looked for in `HOST_OS_XBIN/opam` (deprecated), otherwise it is looked for in `opamHOST_OS_SUFF` (deprecated), otherwise it is expected to be in `PATH`;
+- `dune`, `curl`, `cp` and `ocamlfind` are expected to be in `PATH`.
+
+Using these `DUNE_RELEASE_*` and `HOST_OS_*` environment variables to configurate the path to these binaries is deprecated since `dune-release.1.4.0`, and will no longer be supported in `dune-release.2.0.0`, it is thus recommended to only rely on the `PATH` variable.
```
pitag-ha, line 181:
```suggestion
Using these `DUNE_RELEASE_*` and `HOST_OS_*` environment variables to configure the path to these binaries is deprecated since `dune-release.1.4.0`, and will no longer be supported in `dune-release.2.0.0`, it is thus recommended to only rely on the `PATH` variable.
```

gpetiot: done

README.md
gpetiot: done

gpetiot:
Usually I only add an entry in the changelog when there is a change in the code that impacts the user (so not for design changes, refactoring, etc), whish is not the case for the readme.
In this case I'm not sure since the change of the `git` call is indeed in the code but it's something that was not even documented before so nobody knew of these environment variables unless they read the code. Maybe I should add an entry anyway.

PR #243, opened 2020-06-25
Use long names for curl options and document their usage

gpetiot:
as discussed in #177 

From reading the `curl` documentation it's not obvious to me whether the `--config` option is necessary, does someone has more experience with it?
I don't think we need to add a changelog entry for this PR.

### DIFF
--- a/lib/curl_option.ml
+++ b/lib/curl_option.ml
@@ -1,0 +1,32 @@
+open Astring
+
+type auth = { user : string; token : string }
+
+type t =
+  | Location
+  | User of auth
+  | Silent
+  | Show_error
+  | Config of [ `Stdin | `File of string ]
+  | Dump_header of [ `Ignore | `File of string ]
+  | Data of [ `Data of string | `File of string ]
+  | Data_binary of [ `Data of string | `File of string ]
+  | Header of string
+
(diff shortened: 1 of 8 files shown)


lib/curl_option.ml
```diff
@@ -0,0 +1,32 @@
...
+  | Dump_header of [ `Ignore | `File of string ]
+  | Data of [ `Data of string | `File of string ]
+  | Data_binary of [ `Data of string | `File of string ]
+  | Header of string
+
+let to_string_list opts =
+  List.fold_left
+    (fun acc -> function Location -> "--location" :: acc
+      | User { user; token } -> "--user" :: strf "%s:%s" user token :: acc
+      | Silent -> "--silent" :: acc | Show_error -> "--show-error" :: acc
```
pitag-ha, line 20: Really not important at all, but I was wondering: is it intentional here to have two cases in one line?

gpetiot: It's the result of using the `break-cases= fit` option of ocamlformat (conventional behavior), I agree it doesn't look good here but most of the time it works well

pitag-ha: Cool, thanks for the explanation!

PR #245, opened 2020-07-06
Upgrade to ocamlformat 0.15.0

gpetiot:
Hi, this is what the project would look like after being reformated with the current version (candidate) of ocamlformat 0.15.0.
**Please do not merge until ocamlformat.0.15.0 is available with opam.**

The main changes are due to the improved inconsistency of option `indicate-multiline-delimiters`, a lot of spaces where missing before closing parentheses in the previous versions.
Please let me know if you see any regressions that I may have missed.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,3 +1,3 @@
-version=0.14.2
+version=0.15.0
 profile=conventional
 parse-docstrings=true


bin/cli.ml
```diff
@@ -183,7 +183,7 @@ let handle_error = function
   | Error _ as r ->
       Logs.on_error
         ~pp:(fun fmt (`Msg msg) ->
-          format_linebreaks msg |> Format.pp_print_string fmt)
+          format_linebreaks msg |> Format.pp_print_string fmt )
```
NathanReb, line 186: @gpetiot isn't it weird that it adds a space before the closing parens but not before the opening one?

bin/config.ml
```diff
@@ -16,7 +16,7 @@ let show key =
   | None ->
       let pretty_fields = Config.pretty_fields config in
       StdLabels.List.iter pretty_fields ~f:(fun (key, value) ->
-          Logs.app (fun l -> l "%s: %s" key (show_val value)));
+          Logs.app (fun l -> l "%s: %s" key (show_val value)) );
```
NathanReb, line 19: I also find it disturbing to have that extra space even for inline parenthesized expressions, is that intentional?

gpetiot: The idea is to add the space to distinguish the parentheses that have been opened/closed on the same line from the ones opened/closed in other lines.
There is an exception for `fun`/`function`: the choice has been done for fun and function to only add the space before the closing parenthesis because the space after the opening parenthesis confused the users in this case (and a lot of people think it look bad).

gpetiot:
Trying the last ocamlformat 0.15 RC, the `indicate-multiline-delimiters` changes are postponed until 0.16.

PR #250, opened 2020-07-13
Update the changelog file for the 1.4.0 release

pitag-ha:
This commit adds an entry for the two PR's since release 1.3.3 that have influenced the behavior but didn't have a changelog entry yet. Furthermore, it adapts the changelog to the 1.4.0 release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,12 +2,28 @@
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


pitag-ha:
@NathanReb has just explained me that I have to split this into two parts.

PR #248, opened 2020-07-09
Support creation of draft releases and draft PRs

gpetiot:
~#262 should be merged before~ done

Will fix #246 

- [x] Add a `--draft` option to `dune-release publish` so that the release is created as a draft.
- [x] Add a `--draft` option to `dune-release opam submit` so that the PR is opened as a draft.
- [x] Add a `dune-release undraft` command:
  - [x] writing the release id in a file upon creation so that it can be retrieved while undrafting the release
  - [x] undraft the release on github
  - [x] update the opam file with the new release url
  - [x] writing the pr number in a file upon creation so that it can be retrieved while undrafting the PR
  - [x] undraft the opam-repo PR on github
  - [x] check that we don't try to publish a draft release in a non-draft PR

PR #254, opened 2020-07-21
Remove malfunctional changelog check action

NathanReb:
This is also a great occasion to verify whether it's working or not as we were getting errors on #248 !


NathanReb:
CC @gpetiot it's not working here either, with the same errors. Are you familiar with this action or are you using a different one in ocamlformat?

gpetiot:
I'm not familiar with this at all, in ocamlformat we use a shell script like this: https://github.com/ocaml-ppx/ocamlformat/blob/master/tools/travis-ci.sh#L49

NathanReb:
Ok this one doesn't seem to work at all unfortunately so I'll get rid of it.

I'd prefer using a github action over travis so I'll take a quick look to see how hard it would be to write one myself!

PR #255, opened 2020-07-21
Fix a couple bugs in opam pkg subcommand with --dist-file and --dist-uri

NathanReb:
The option used to have lower priority than the recently introduced `url_file` which made it impossible to overwrite it if you wanted to publish the tarball to github and somewhere else but wanted to use the URL for the second one in opam.

I think it's a side effect of the `Pkg` module which contains more than it really should. We should try to fix this as we go.
Actually now that I think of it I'll add a commit that gets rid of the corresponding field in `Pkg.t` since it's not used anymore.


NathanReb:
Good point, I wanted to do it and then forgot, I'll do that!

NathanReb:
While adding the test I found another bug so I added a test and a fix for this one as well, let me know what you think!

PR #256, opened 2020-07-22
Test release process for non-github URIs

gpetiot:
One test where the publication of the doc fails, one where the publication of the distrib fails.

### DIFF
--- a/tests/bin/non-github-doc-uri/run.t
+++ b/tests/bin/non-github-doc-uri/run.t
@@ -1,0 +1,107 @@
+We need a basic opam project skeleton
+
+    $ cat > CHANGES.md << EOF \
+    > ## 0.1.0\
+    > \
+    > - Some other feature\
+    > \
+    > ## 0.0.0\
+    > \
+    > - Some feature\
+    > EOF
+    $ cat > whatever.opam << EOF \
+    > opam-version: "2.0"\
+    > homepage: "https://whatever.io"\
+    > dev-repo: "git+https://whatever.io/dev/whatever.git"\
+    > doc: "https://whatever.io/doc/main.html"\
+    > description: "whatever"\
+    > EOF
+    $ touch README
+    $ touch LICENSE
+    $ cat > dune-project << EOF \
+    > (lang dune 2.4)\
+    > (name whatever)\
+    > EOF
+
+We need to set up a git project for dune-release to work properly
+
+    $ git init > /dev/null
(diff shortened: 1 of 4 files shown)


gpetiot:
@NathanReb @pitag-ha is there other tests that we should add in this PR that I didn't think of by the way? (not necessarily about non-github uris)

tests/bin/non-github-doc-uri/run.t
```diff
@@ -0,0 +1,162 @@
...
+    
+    [+] Distribution for whatever 0.1.0
+    [+] Commit ...
+    [+] Archive _build/whatever-0.1.0.tbz
+
+We do the whole dune-release process
+
+(1) distrib
+
+    $  yes | dune-release distrib --dry-run
```
NathanReb, line 94: This appears twice I think!

tests/bin/non-github-uri/run.t
```diff
@@ -0,0 +1,153 @@
...
+    $ git add CHANGES.md whatever.opam dune-project README LICENSE
+    $ git commit -m "Initial commit" > /dev/null
+    $ dune-release tag -y
+    [-] Extracting tag from first entry in CHANGES.md
+    [-] Using tag "0.1.0"
+    [+] Tagged HEAD with version 0.1.0
+
+We make a dry-run release
+
+    $ dune-release distrib --dry-run
```
NathanReb, line 37: Same here, this is duplicated below!

tests/bin/non-github-doc-uri/run.t
```diff
@@ -0,0 +1,162 @@
...
+    [ OK ] package tests
+    -: rmdir _build/whatever-0.1.0
+    
+    [+] Distribution for whatever 0.1.0
+    [+] Commit ...
+    [+] Archive _build/whatever-0.1.0.tbz
+
+(2) publish
+
+    $  yes | dune-release publish --dry-run
```
NathanReb, line 149: How about making this `dune-release publish doc --dry-run` so that the test don't break if we ever decide to change the order of publishing?

gpetiot: done

tests/bin/non-github-uri/run.t
gpetiot: done

tests/bin/non-github-doc-uri/run.t
gpetiot: done

PR #257, opened 2020-07-23
Normalize git comands calls and error reporting

### LINKED ISSUE #253: For better error reporting: use `Sos.run_out_err` more
For example, in `github.ml` there are 3 functions that use`Sos.quite_run` or `Sos.run` where I think they could use `Vcs.run_git`. The latter provides a considerably better error reporting. So I think it might be good to change that. I've noticed when I ran into the problem that's already described in issue #212 (not on MacOs though) and received an error message that was not helpful.

I made the title more general, since I think it might be good to have a quick look through the code where else we could improve the error reporting easily by using `Sos.run_out_err` instead of `Sos.run_gen`, either directly or through `Vcs.run_git`.

gpetiot:
Fix #253, no diff on the tests so far, maybe we need to add more.


NathanReb (commented):
Looks good, thanks!

This needs a changelog entry though as users will get feedback now when some of those git commands fail!

PR #258, opened 2020-07-27
Add @emillon's changelog check action

NathanReb:
This PR adds a github action to check that the changelog has been updated unless the `no changelog` label is added.

Thanks to @emillon for working on this!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,20 @@
+name: Changelog check
+
+on:
+  pull_request:
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


NathanReb:
It seems to work as intended!

NathanReb:
@gpetiot @pitag-ha could you please have a look at this so that we can merge it quick and make sure the changelog never goes out of date again?

I think the travis failures are completely unrelated so we should just ignore them. I'm looking how to restart the builds but I don't seem to have sufficient permissions for that atm!

PR #262, opened 2020-08-04
Allow the dry-run mode to continue even after some API call's response were expected by using placeholder values

gpetiot:


CHANGES.md
```diff
@@ -18,6 +18,7 @@
   a custom distrib archive file. (#255, @NathanReb)
 - Use int64 for timestamps. (#261, @gpetiot)
 - Define the order of packages (#263, @gpetiot)
+- Define default values for dry-run mode (#262, @gpetiot)
```
NathanReb, line 21: That's not very helpful to users, I think it should instead mention how it improves the dry-run mode by allowing it to continue even after some API call's response were expected by using placeholder values.

PR #263, opened 2020-08-04
Define the order of packages

### LINKED ISSUE #249: Sort package names alphabetically in Opam PR
When releasing multiple packages at once, `dune-release` seems to pick an arbitrary order for the package names in the release. For instance:

```
[new release] alcotest-lwt, alcotest, alcotest-async and alcotest-mirage (1.2.0)
Open • CraigFe wants to merge 1 commit into master from release-alcotest-1.2.0


  Lwt-based helpers for Alcotest

  • Project page: https://github.com/mirage/alcotest
  • Documentation: https://mirage.github.io/alcotest

  ##### CHANGES:
```

When it does this, it constructs the PR description from the first package in the list. It'd be good to have some way of giving `dune-release` either a file from which to take the description (à la [`hub`](https://hub.github.com/) `pull-request -F <FILE>`), or to tell it that one particular package is the "canonical" one and should always appear first. (The latter would suffice for all of the projects I regularly release.)

gpetiot:
this is making the tests more deterministic
also fix #249

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@
   the regular dune-release generated archive to be around even though you specified
   a custom distrib archive file. (#255, @NathanReb)
 - Use int64 for timestamps. (#261, @gpetiot)
+- Define the order of packages (#263, @gpetiot)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


gpetiot:
I think it's more of an unspoken rule, I didn't find anything about this naming convention anywhere. In `dune-release` the package order only determines the linting order, we are not relying on that for anything significant as far as I'm aware of.

PR #264, opened 2020-08-07
Sanitize usage of distrib_uri

gpetiot:
Should not change the behavior of dune-release:
- remove `distrib_uri` field from `Pkg.t` type
- remove `?distrib_uri` arg from `Pkg.v` function
- remove `?raw` arg from `Pkg.distrib_uri` function
- ~add `?uri` arg to `Pkg.distrib_uri` function~
- ~add `?uri` arg to `Pkg.distrib_user_and_repo` function~
- pass `?distrib_uri` arg to functions calling one of these two functions
- ~split `Pkg.distrib_uri` into `Pkg.infer_distrib_uri` and `Pkg.normalize_distrib_uri`~
- rename `Pkg.distrib_uri` to `Pkg.infer_distrib_uri`

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -44,7 +44,6 @@ type t = {
   change_logs : Fpath.t list option;
   licenses : Fpath.t list option;
   distrib : Distrib.t;
-  distrib_uri : string option;
   distrib_file : Fpath.t option;
   publish_msg : string option;
 }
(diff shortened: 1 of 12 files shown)


lib/pkg.ml
```diff
@@ -262,32 +261,28 @@ let distrib_uri_of_homepage p =
...
-    | None -> (
-        distrib_uri_of_homepage p >>= function
-        | Some u -> Ok u
-        | None -> (
-            distrib_uri_of_dev_repo p >>= function
-            | Some u -> Ok u
-            | None -> err_not_found () ) )
-  in
-  uri >>= fun uri ->
+let distrib_uri ?uri p =
```
NathanReb, line 264: Just to confirm, this extra argument will go away once we move this function from `Pkg` to `Pkg.Github` right?

NathanReb: Also it's probably better written directly at the call site as a `Option.value_lazy ~default:(fun () -> Pkg.distrib_uri pkg) distrib_uri`.

NathanReb:
Also now that I think about it, we have to be careful about this as it should mainly come from the url_file (as done in `dune-release opam pkg`) and fallback to `Pkg.distrib_uri` otherwise and I'm not sure that's what we've been doing everywhere so far.

That's independent from this PR but we have to look into it!

NathanReb:
The function should probably be called `infer_distrib_uri` to clarify what it does.

lib/pkg.ml
gpetiot: For now this argument corresponds to the value of the `--distrib-uri` option, but I can rewrite the check fir this option earlier indeed.

gpetiot: I removed the argument, let me know what you think

gpetiot:
> I'm not sure splitting it into two functions is what we need here, in particular I don't think there's a point in normalizing the user provided URI when passing it to the delegate. I understand this preserves the old behaviour but I think it also emphasizes how broken it was.

Agreed I kept the normalization part for the inferred distrib_uri only.

NathanReb (approved):

PR #267, opened 2020-08-18
Infer the main package from the dune-project if possible

### LINKED ISSUE #265: dune-release opam submit uses synopsis of a random opam package
When I `dune-release opam submit` in the ocaml-tls (https://github.com/mirleft/ocaml-tls) repository, the generated PR contains the synopsis of the `tls-mirage` opam package, though I'd prefer it to contain the contents from the `tls` opam package. see https://github.com/ocaml/opam-repository/pull/16984

I think there's a general issue with multiple opam packages in a single repository (see #244). I'd suggest to denote a "main" opam package (and as I understand, this is the one without `-`: `irmin` in the https://github.com/mirage/irmin repository, `dns` in https://github.com/mirage/ocaml-dns, `dune` in https://github.com/ocaml/dune, `tls` in https://github.com/mirleft/ocaml-tls) which content (synopsis, homepage, etc.) is used where needed.

I know #249 has been merged which likely modifies the behaviour slightly (though not clear to me whether this is good now). I used dune-release 1.4.0 here.

gpetiot:
fix #265
~This PR defines a new package `Mlist` in `Stdext`, for "main" list (didn't find a better name) to distinguish the main element of the list from others elements of the list, and we maintain this distinction for packages everywhere it makes sense. So the main package can be retrieved directly, without assuming it is the first element of the list or that it has the shortest name of all elements.~

edit: I improved the PR to define `Pkg.infer_main_pkg` instead (in the same way the new `Pkg.infer_pkg_names` was defined), so the main package can be passed along the package list, parameters have also been named to avoid being confused by generic pkg/pkgs variables. Let me know what you think.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,7 @@
   dune-project would be silently ignored by `dune-release distrib`.
 - Fix rewriting of github references in changelog (#330, @gpetiot)
 - Fixes a bug under cygwin where dune-release was unable to find the commit hash corresponding to the release tag (#329, @gpetiot)
+- Infer the main package from the dune-project if possible (#267, @gpetiot)
 
 ### Security
 
--- a/lib/opam.mli
+++ b/lib/opam.mli
@@ -22,11 +22,13 @@ val prepare :
   remote_repo:string ->
   opam_repo:string * string ->
   version:string ->
-  string list ->
+  main_pkg:string ->
+  all_pkgs:string list ->
   (string, R.msg) result
-(** [prepare ~local_repo ~version pkgs] adds the packages [pkg.version] to a new
-    branch in the local opam repository [local_repo], using the commit message
-    [msg] (if any). Return the new branch. *)
+(** [prepare ~local_repo ~version ~main_pkg ~all_pkgs] adds the packages from
+    [all_pkgs] to a new branch named "release-[main_pkg]-version" in the local
+    opam repository [local_repo], using the commit message [msg] (if any).
+    Commits and pushes the changes, then returns the new branch. *)
 
 (** {1:file Files} *)
 
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -438,6 +438,10 @@ let infer_name dir =
                      (name <name>) to dune-project");
               exit 1))
 
+let infer_main_pkg dir pkgs =
+  infer_name dir >>| fun main_pkg_name ->
+  List.find (fun pkg -> String.equal pkg.name main_pkg_name) pkgs
+
 let v ~dry_run ?name ?version ?tag ?(keep_v = false) ?delegate ?build_dir
     ?opam:opam_file ?opam_descr ?readme ?change_log ?license ?distrib_file
     ?publish_msg () =
--- a/bin/opam.ml
+++ b/bin/opam.ml
@@ -160,7 +160,7 @@ let open_pr ~dry_run ~changes ~remote_repo ~user ~distrib_user ~branch ~token
         | Error _ -> msg ())
 
 let submit ?distrib_uri ~token ~dry_run ~yes ~opam_repo ~user local_repo
-    remote_repo pkgs auto_open =
+    remote_repo ~main_pkg:pkg ~pkgs auto_open =
   List.fold_left
     (fun acc pkg ->
       get_pkg_dir pkg >>= fun pkg_dir ->
@@ -175,7 +175,6 @@ let submit ?distrib_uri ~token ~dry_run ~yes ~opam_repo ~user local_repo
           Ok 1)
     (Ok 0) pkgs
   >>= fun _ ->
-  let pkg = List.hd pkgs in
   Pkg.version pkg >>= fun version ->
   list_map Pkg.name pkgs >>= fun names ->
   let title = strf "[new release] %a (%s)" (pp_list Fmt.string) names version in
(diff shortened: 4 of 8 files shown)


pitag-ha (commented):
I like the idea of making use of `infer_name` to detect the main package and storing it in your new data type a lot! I just think doing that everywhere, also in places where it's not important what the main package is, can be a bit confusing (see comment).

NathanReb (commented):
Looks great! I think we indeed need to clarify this part in the code!

I have a couple minor comments but other than that it's :+1: 

Although it's unrelated to this PR, I'm wondering if we shouldn't have checks to make sure the list of packages is consistent e.g.:
- that the main package we inferred does have an opam file at the root
- that all packages explicitly passed do have an opam file at the root

Doing those checks early would probably make for better error reporting.

I also agree with @pitag-ha that not all places where we need a list of packages necessarily need a main package. For instance, what happens when you want to release a subset of your packages, none of which is the main one?

Might be worth clearly identifying how this main package is used so we can refine all this in 2.0.0.

CHANGES.md
```diff
@@ -31,6 +31,7 @@
   (#266, @NathanReb)
 - Improve trimming of the changelog to preserve the indentation of the list of changes. (#268, @gpetiot)
 - Trim the data of the `url` file before filling the `url.src` field. This fixes an issue that caused the `url.src` field to be a multi-line string instead of single line. (#270, @gpetiot)
+- Handle the main package consistently (#267, @gpetiot)
```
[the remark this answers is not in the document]
gpetiot: if fixes #265 where we assume the package with the shortest name is the main name, I will improve the changelog

lib/stdext.ml
```diff
@@ -68,3 +68,9 @@ module List = struct
     in
     fmap [] l
 end
+
+module Mlist = struct
```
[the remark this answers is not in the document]
gpetiot: Good idea, done!

lib/pkg.ml
```diff
@@ -440,6 +440,14 @@ let infer_name dir =
                   m "cannot determine name automatically: use `-p <name>`");
               exit 1 ) )
 
+(* [infer_pkg_names] initially defined above cannot call [infer_name] or there
+   will be a cyclic dependency. *)
+let infer_pkg_names dir pkgs =
```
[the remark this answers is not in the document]
gpetiot: I modified this to now pass the main package or all packages where it makes sense

lib/pkg.ml
```diff
@@ -440,6 +440,14 @@ let infer_name dir =
...
+  infer_pkg_names dir pkgs >>= fun pkgs ->
```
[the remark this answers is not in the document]
gpetiot: I changed this and defined an `infer_main_pkg` instead

lib/opam.mli
```diff
@@ -22,11 +22,12 @@ val prepare :
...
-  string list ->
+  main_pkg:string ->
+  all_pkgs:string list ->
   (string, R.msg) result
-(** [prepare ~local_repo ~version pkgs] adds the packages [pkg.version] to a new
-    branch in the local opam repository [local_repo], using the commit message
-    [msg] (if any). Return the new branch. *)
+(** [prepare ~local_repo ~version ~main_pkg ~all_pkgs] adds the packages
+    [pkg.version] to a new branch in the local opam repository [local_repo],
+    using the commit message [msg] (if any). Return the new branch. *)
```
pitag-ha, line 30: I don't quite understand this documentation (although, to be honest, I also didn't understand the old one). What is `[pkg.version]`? And what does the whole sentence mean? Does it mean that the packages from `[all_pkgs]` are added to a new branch named `release-[main_package]-[version]` in `[local_repo]`? And, if so, doesn't the function do a bit more, such as also committing and pushing that change?

lib/pkg.ml
```diff
@@ -436,6 +436,10 @@ let infer_name dir =
                   m "cannot determine name automatically: use `-p <name>`");
               exit 1 ) )
 
+let infer_main_pkg dir pkgs =
+  infer_name dir >>| fun main_pkg_name ->
+  List.find (fun pkg -> String.equal pkg.name main_pkg_name) pkgs
```
pitag-ha, line 441: I'm wondering if there might be projects where that strategy might fail. I think it's very unlikely. But even being very unlikely, do you think it might make sense to add a lint check that checks if the main package can be inferred (in a separate PR)?

CHANGES.md
```diff
@@ -40,6 +40,7 @@
 - Trim the data of the `url` file before filling the `url.src` field. This fixes an issue that caused the `url.src` field to be a multi-line string instead of single line. (#270, @gpetiot)
 - Fix a bug causing dune-release to exclude all hidden files and folders (starting with `.`) at the
   repository from the distrib archive (#298, @NathanReb)
+- Do not assume the main package is the one with the shortest name (#267, @gpetiot)
```
pitag-ha, line 43: Is that what we were doing before? I thought what we were doing before was in some places (`prepare`) to take the package with the shortest name, but in other places (`submit`) even just the first package in the list (leading to the problem described in #265 of taking the synopsis of `tls-mirage` instead of the one from `tls` for the PR message). I think what you are doing now is to infer the main package from the dune-project; if that fails, take the one with the shortest name, if that's unique, and if it's not unique, infer it from the README, no? Maybe you can put the sentence without negation, something like "Infer the main package from the dune-project, if possible".

lib/opam.mli
gpetiot: Indeed! I will update the documentation

lib/pkg.ml
gpetiot: It makes a lot of sense indeed!

CHANGES.md
gpetiot: That's a better changelog entry indeed, that should be less confusing to users

pitag-ha (approved):

gpetiot:
Diverged too much

PR #268, opened 2020-08-18
Improve trimming of the changelog

### LINKED ISSUE #260: Changelog misformatted
The function `publish_msg` uses the changelog if the publish message is unavailable.  It will trim the text of this entry, which can cause problems if the entry uses unordered lists of changes.

E.g. instead of:
  * Change A
  * Change B

You get:
* Change A
  * Change B

I don't know what the reason for trimming is, but maybe it's unnecessary?  Or you may need to also trim the same amount of spaces as for the first line after each `'\n'` for consistency.

gpetiot:
fix #260

### DIFF
--- a/tests/bin/no_doc/run.t
+++ b/tests/bin/no_doc/run.t
@@ -3,7 +3,8 @@ We need a basic opam project skeleton
     $ cat > CHANGES.md << EOF \
     > ## 0.1.0\
     > \
-    > - Some other feature\
+    >  - Change A   \
+    >  - Change B\
     > \
     > ## 0.0.0\
     > \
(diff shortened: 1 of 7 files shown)


tests/bin/no_doc/run.t
```diff
@@ -3,7 +3,8 @@ We need a basic opam project skeleton
     $ cat > CHANGES.md << EOF \
     > ## 0.1.0\
     > \
-    > - Some other feature\
+    >  - Change A\
+    >  - Change B\
```
NathanReb, line 7: Maybe we should add trailing whitespaces here to make sure they're being dropped!

Although I guess this is also the kind of things we could extract a function for and have finer grained unit tests. What do you think?

NathanReb (commented):
Looks good! Thanks for taking care of his!

I have a suggestion regarding testing that I think would improve this but it's not strictly necessary!

tests/bin/no_doc/run.t
gpetiot: Good idea, done.

NathanReb (approved):

PR #269, opened 2020-08-18
Linting will check synopsis field when there is no descr field

### LINKED ISSUE #259: dune-release demands a `description` field even when a `synopsis` one is present
on https://github.com/c-cube/qcheck/commit/6a170e34269164d45aba6303dff1c77cfb8b8fd0 I get errors when running `dune-release lint` because it seems to simply ignore the synopsis field. `opam lint` is all green so why is `dune-release` more strict?

gpetiot:
fix #259

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@
 
 - Attach the changelog to the annotated tag message (#283, @gpetiot)
 - Deprecate the use of delegates in `dune-release publish` (#276, @pitag-ha)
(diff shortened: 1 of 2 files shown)


gpetiot:
Can we merge this one?

NathanReb:
If we're going to change that check to this we might as well remove it since that's already enforced by `opam lint` as far as I can tell.
To be honest I'm not sure what to do with this for now, I'd rather wait on the 2.0 lint changes, linting can always be disabled until then so this doesn't have very high priority.

gpetiot:
Postponed until linting rework

PR #272, opened 2020-08-25
Try to publish the release asset again after it failed

### LINKED ISSUE #211: dune-release publish sometimes fails
```
[?] Upload _build/mirage-entropy-v0.5.1.tbz as release asset? [Y/n]
[-] Uploading _build/mirage-entropy-v0.5.1.tbz as a release asset for v0.5.1 via github's API
dune-release: [ERROR] Could not extract archive url from:
                       {"message":"Error saving
                      asset","request_id":"D9BE:35EF:3D95:4F50:5E5401F7"}
```

And now the release is created on github, but the tarball is not uploaded. A `dune-release publish distrib` fails since it attempts to re-create the release.

To workaround, I delete the release and start `dune-release publish` again. Could there be a better way to deal with these errors? Could there be a more robust way to upload the tarball?

I suggest a `dune-release publish` subcommand which only uploads the tarball would be a good first step.

gpetiot:
Fix #211

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 - Add a `--token` option to `dune-release publish` and `dune-release opam` commands to specify a github token. This allows dune-release to be called through a Github Actions workflow and use the github token through an environment variable. (#284, @gpetiot)
 - Log curl calls on verbose/debug mode (#281, @gpetiot)
+- Try to publish the release asset again after it failed (#272, @gpetiot)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


NathanReb (commented):
I'm not sure this is entirely the right fix, there are two main reasons here:
1. Prompting for retry is a good idea but we're doing so without telling the suer about the error that occurred so they have no idea whether retrying has any chance of succeeding or not. I'd suggest we at least print out the error first. Also note that passing the `--yes` option here can cause a infinite loop if the error for the upload is legit! We should allow a maximum number of retries.
2. Imagine the user exits by answering `no` at some point, they end up in the exact same situation than before, the release has already been created and they have to delete it again. This makes me reconsider @kit-ty-kate recent PRs a bit. We can discuss the proper solution here at today's meeting!

gpetiot:
I've added a limit for the number of retries and the error message is now printed before retrying.

NathanReb (approved):
Looks good!

PR #275, opened 2020-09-01
Add support for lightweight tags

kit-ty-kate:
Currently dune-release does not support lightweight tags (aka. default tags), so for instance if a user already pushed a tag in the following manner:
```
$ git tag v0.1
$ git push origin master
$ git push origin v0.1
```
dune-release will not realize that the tag already exists and will want to create a nonsensical tag instead (e.g. `v0.1-16-g21abe5d`).

Currently https://github.com/ocamllabs/dune-release/pull/219 is needed to be able to use `dune-release` directly afterward, but it's still makes sense to allow lightweight tags.

### DIFF
--- a/lib/vcs.ml
+++ b/lib/vcs.ml
@@ -96,7 +96,7 @@ let git_describe ~dirty r commit_ish =
   let dirty = dirty && commit_ish = "HEAD" in
   let git_describe =
     Cmd.(
-      git_work_tree r % "describe" % "--always"
+      git_work_tree r % "describe" % "--always" % "--tags"
       %% on dirty (v "--dirty")
       %% on (not dirty) (v commit_ish))
   in


NathanReb:
I understand it can be annoying but given `dune subst` also requires annotated tag we can't suppor tlightweight tag in dune-release at the moment.

To be honest I'm not sure we even should but if we ever have to support them we need a release of dune that does first and we'd also need to drop support for older dune versions so it's not going to happen anytime soon.

For now users will have to use annotated tags to be compatible with dune-release which is what they get from a `dune-release tag` invocation. I understand one might want to go through the tagging themselves but creating an annotated one seems like a decent contract. I'll update the doc to reflect that.

I am closing this, feel free to open an issue if you want to discuss other forms of support for lightweight tags!

Issue #188, opened 2019-11-22
Replacing delegates


NathanReb:
The idea is to offer non github-based workflows an alternative way to cooperate with `dune-release` that requires minimal maintenance on our side and to simplify the most common case, publishing to github, in the process.

## The issue

Currently delegates can be used to:
- publish documentation
- publish tarballs
- publish arbitrary things or "alt-" targets

The way it works is that `dune-release` tries to infer whether it should publish to github (the default and so far most common workflow) or publish using a delegate by parsing URLs in the opam file.
These parts of the code are fragile, hard to maintain and lead to bad error reporting at the moment.

## Our proposal

The idea here would be to remove the delegate API and the corresponding code in `dune-release` so that it natively only supports github publication and to offer other users a way to plug their scripts into the `dune-release` workflow by ensuring intermediate `dune-release` commands produce the right artifacts to pass on to those scripts rather than invoking them ourselves.

Currently delegates are only passed the path to the distribution tarball, the path to the generated documentation directory and the "publication message". The first two are already written to the file system to a sub directory of `_build` in a fairly consistent manner. The publication message isn't but it would make sense for dune-release to do so as it is reused across different commands and re-computed by parsing the changelog everytime it's required.

What I suggest here is to add a simple `delegate-info` command that prints to the standard output the paths and values currently passed to delegate commands. That way external publication tools can use that to get the information they need and use to get from the command line when invoked by `dune-release`. This allows us to easily maintain a minimal interface with alternative publication workflows while allowing us to change the internal behaviour of tools.

A last point worth mentioning is that to open PRs to opam-repository, dune-release needs to know where the tarball can be downloaded from. This can be set using the `--distrib-uri` option of `dune-release opam`. Although this option priority is currently broken, once it is fixed it should provide a way for delegates to give `dune-release` the information it needs to take on with the rest of the release process.

In short, the new proposed alternative is to let external release scripts invoke dune-release rather than the other way around which seems like a better approach. This plan not only is better for dune-release maintenance but will also make it easier to extend the `delegate-info` API if external tools need more information from `dune-release` so I believe it's better for everyone involved.

I suggest the following course of action:
1. Submit this plan to the community to give affected users an opportunity to speak up and make some adjustments if needed
2. Add the `delegate-info` command and release it in a minor release
3. Deprecate all other delegate related features in a subsequent minor release
4. Completely drop the old delegate API in the next major release

NathanReb:
[...]

PR #276, opened 2020-09-01
Deprecate delegates

pitag-ha:
This commit deprecates the use of delegates and publishing artefacts other than the distribution archive or the documentation, since those functionalities will be removed in the next major release. Deprecating delegates is point 3. of #188.

In order to remove delegates, we'll have to 
- remove the modules `Deprecate_delegates` and `Delegate`
- follow the provoked compile errors
- get rid of the delegate structure in `Pkg` 
- think about where to check if the user is using something else than github (provided in the opam `doc:` or `homepage:` field) to let them know early that dune-release doesn't support that. Maybe through the linter?

### DIFF
--- a/bin/publish.ml
+++ b/bin/publish.ml
@@ -42,7 +42,9 @@ let publish_doc ~specific ~dry_run ~yes pkg_names pkg =
   | _ when specific -> publish_doc ~dry_run ~yes pkg_names pkg
   | Error _ | Ok "" -> (
       match Pkg.delegate pkg with
-      | Ok (Some _) -> publish_doc ~dry_run ~yes pkg_names pkg
+      | Ok (Some _) ->
+          App_log.unhappy (fun l -> l Deprecate_delegates.warning_usage);
+          publish_doc ~dry_run ~yes pkg_names pkg
       | Error _ | Ok None ->
           Pkg.name pkg >>= fun name ->
           Pkg.opam pkg >>= fun opam ->
@@ -59,12 +61,6 @@ let publish_distrib ?token ?distrib_uri ~dry_run ~yes pkg =
   Pkg.distrib_file ~dry_run pkg >>= fun archive ->
   Pkg.publish_msg pkg >>= fun msg ->
   Delegate.publish_distrib ?token ?distrib_uri ~dry_run ~yes pkg ~msg ~archive
-
-let publish_alt ?distrib_uri ~dry_run pkg kind =
-  App_log.status (fun l -> l "Publishing %s" kind);
-  Pkg.distrib_file ~dry_run pkg >>= fun archive ->
-  Pkg.publish_msg pkg >>= fun msg ->
-  Delegate.publish_alt ?distrib_uri ~dry_run pkg ~kind ~msg ~archive
 
 let publish ?build_dir ?opam ?delegate ?change_log ?distrib_uri ?distrib_file
     ?publish_msg ?token ~name ~pkg_names ~version ~tag ~keep_v ~dry_run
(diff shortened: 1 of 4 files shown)


gpetiot:
The `delegate` field in Pkg is only set when calling `Pkg.v` from `bin/publish.ml` but it can be directly passed to the delegate calls from this same file as it is the only one using delegates anyway. I can do that in another PR in the meantime if it sounds good.

gpetiot (approved):
It looks good to me, I wish we could remove them soon!

pitag-ha:
Yes, I totally agree that it would be great to remove them soon...

bin/publish.ml
```diff
@@ -157,8 +156,9 @@ let man =
     `P "$(mname) $(tname) [$(i,OPTION)]... [$(i,ARTEFACT)]...";
     `S Manpage.s_description;
     `P
-      "The $(tname) command publishes package distribution archives and other \
-       artefacts.";
+      ( Deprecate_delegates.artefacts_warning
```
NathanReb, line 159: A separate paragraph might be better here!

bin/publish.ml
```diff
@@ -120,13 +118,15 @@ let artefacts =
...
+        | kind ->
+            App_log.unhappy (fun l -> l Deprecate_delegates.warning_usage);
+            `Ok (`Alt kind) )
     | s -> `Error (strf "`%s' unknown publication artefact" s)
   in
   let printer ppf = function
     | `Doc -> Fmt.string ppf "doc"
     | `Distrib -> Fmt.string ppf "distrib"
-    | `Alt a -> Fmt.pf ppf "alt-%s" a
+    | `Alt a -> Fmt.pf ppf Deprecate_delegates.alt_artefacts_pp a
```
NathanReb, line 129: This is just used by Cmdliner to report errors I think, not sure it's worth redifining the printer.

bin/publish.ml
NathanReb: Hmm ok maybe not but what would make sense it to move it after the text below so that we have at least mentioned the alternative artifacts!

bin/publish.ml
```diff
@@ -85,7 +81,7 @@ let publish ?build_dir ?opam ?delegate ?change_log ?distrib_uri ?distrib_file
     match artefact with
     | `Doc -> publish_doc ~specific:specific_doc ~dry_run ~yes pkg_names pkg
     | `Distrib -> publish_distrib ?token ?distrib_uri ~dry_run ~yes pkg
-    | `Alt kind -> publish_alt ?distrib_uri ~dry_run pkg kind
+    | `Alt kind -> Deprecate_delegates.publish_alt ~dry_run pkg kind
```
NathanReb, line 84: I think it makes more sense to just add a log, something like:
```suggestion
    | `Alt kind ->
      Deprecate.Delegate.artifact ();
      publish_alt ?distrib_uri ~dry_run pkg kind
```
Where `Deprecate.Delegate.artifact` would just log the right warning.

lib/deprecate_delegates.ml
```diff
@@ -0,0 +1,37 @@
+let warning =
+  "dune-release delegates are deprecated. They will be removed in version \
+   2.0.0."
```
NathanReb, line 3: I think we should add pointers to the replacement solution in the warnings. Somehting like:

"You should write scripts invoking dune-release instead. Note that if you require some values that used to be passed by dune-release to your delegate you can obtain them with the `dune-release delegate-info` command added in 1.4.0"

I'm not sure if it should appear everywhere but at least for all the big entry points so the doc and the logs I guess.

NathanReb (commented):
I'm slightly late for this sry.

This looks good but I think it could be slightly simplified.

For instance a lot of the strings in the `Deprecate_delegates` module wrap the text they were supposed to be added to instead of just concatenating them where this doc text is defined.

A generic `Deprecate` module would be better here so we could have all deprecation related features there and remove or empty the module when we release 2.0. For instance the opam 1.2 deprecation should also go there, eventually in different submodules.

bin/publish.ml
pitag-ha: Yes, that's better. I've moved it below and I've also moved all warnings behind the concept they are for.

bin/publish.ml
pitag-ha: The idea was to make removing delegates easier: when removing delegates we'd get a compilation error here when removing the `Deprecate_delegates` warning and wouldn't forget about this match arm. But I've reverted this.

bin/publish.ml
pitag-ha: Ok, I've done that now.

lib/deprecate_delegates.ml
pitag-ha: Yes, that's a good idea. I've done that.

pitag-ha:
Thanks, @NathanReb ! I've implemented those changes in PR #302.

PR #277, opened 2020-09-01
Create github releases only if necessary

kit-ty-kate:
This PR is built on top of https://github.com/ocamllabs/dune-release/pull/219 by @julow which allows for git tags to not be created if they already exist. This PR takes a notch further and does not create a github release if it already exists.

This might be useful for people who want custom github releases messages or something and still want to use dune-release for the rest. It might also be useful for people who, like myself, are too used to `opam-publish` and already created a tag and release before calling `dune-release`.
People might also be in the same situation if something did not work when they used `opam-publish`, then tries to use `dune-release` instead but the workflow is too different for them and they give up (either give up using opam altogether or release by hand).

I hope this can help people transition from opam-publish if they want to or simply allow people to have finer control over what they are doing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 - Don't try to push the tag if it is already present and point to the same ref on the remote.
   `dune-release` must guess which URI to pass to `git push` and may guess it wrong.
   This change allows users to push the tag manually to avoid using that code. (#219, @Julow)
+- Don't try to create the release if it is already present and points to the same tag (#277, @kit-ty-kate)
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
(successfully tested together with #219 and #275 to create https://github.com/ocaml/opam-repository/pull/17100)

pitag-ha:
Thanks, that's a nice improvement! I've just rebased. Could you write a changelog entry? Apart from that it looks good to me.

kit-ty-kate:
Done

pitag-ha:
Thanks!

PR #278, opened 2020-09-02
Mention that dune-release only supports annotated tags

NathanReb:
This confused users in the past so it'd be better to clarify it in the README.


gpetiot (approved):
Looks good, thanks.

PR #281, opened 2020-09-03
Log curl calls on debug

### LINKED ISSUE #280: `-vvv` does not log `curl` calls
Currently `-vvv` does not log the `curl` requests; that makes debugging quite difficult when it comes to GitHub API related failures. 
Perhaps logging should also happen there?

gpetiot:
Fix #280

No diff in the tests since we only use dry-mode to not submit pull requests and so on.

### DIFF
--- a/lib/github.ml
+++ b/lib/github.ml
@@ -178,8 +178,16 @@ let run_with_auth ?(default_body = `Null) ~dry_run ~auth curl_t =
     >>| fun () -> default_body
   else
     OS.Cmd.must_exist (Cmd.v "curl") >>= fun _ ->
-    match Curly.(run ~args (Request.make ~url ~meth:`POST ())) with
-    | Ok Curly.Response.{ body; _ } -> Json.from_string body
+    let req = Curly.Request.make ~url ~meth:`POST () in
+    Logs.debug (fun l ->
+        l "[curl] executing request:@;<1 2>%a" Curly.Request.pp req);
+    Logs.debug (fun l ->
+        l "[curl] with args:@;<1 2>%a" (Fmt.list ~sep:Fmt.sp Fmt.string) args);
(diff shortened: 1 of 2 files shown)


lib/github.ml
```diff
@@ -178,8 +178,16 @@ let run_with_auth ?(default_body = `Null) ~dry_run ~auth curl_t =
     >>| fun () -> default_body
   else
     OS.Cmd.must_exist (Cmd.v "curl") >>= fun _ ->
-    match Curly.(run ~args (Request.make ~url ~meth:`POST ())) with
-    | Ok Curly.Response.{ body; _ } -> Json.from_string body
+    let req = Curly.Request.make ~url ~meth:`POST () in
+    Logs.debug (fun l ->
+        l "[curl] executing request:@;<1 2>%a" Curly.Request.pp req);
+    Logs.debug (fun l ->
+        l "[curl] with args:@;<1 2>%a" (Fmt.list Fmt.string) args);
```
NathanReb, line 185: Might be good to set a `~sep` there don't you think? I'm not sure what it defaults to!

NathanReb (commented):
Thanks for working on this!

I have one minor comment but otherwise it looks good!

It's probably worth a changelog entry since now we can use the verbose output to get more information on curl failures so that's noticeable changes from the user perspective as well!

NathanReb (commented):
Thanks for working on this!


I have one minor comment but otherwise it looks good!

It's probably worth a changelog entry since now we can use the verbose output to get more information on curl failures so that's noticeable changes from the user perspective as well!

NathanReb:
Sorry about the double review, github kind of failed me here!

lib/github.ml
gpetiot: indeed, done!

PR #283, opened 2020-09-07
Attach the changelog to the annotated tag message

### LINKED ISSUE #274: Feature request: add changelog to annotated tag message
Hi,

Annotated tags (the kind `dune-release` uses) have messages attached to them (like commits).

At the moment the string `"Distribution x.y.z"` is used (can be overridden on the commandline).
I'm thinking it could contain the changelog section corresponding to this release (the one that's used in the github release and opam PR).

How does that sound?

gpetiot:
Fix #274

### DIFF
--- a/bin/tag.ml
+++ b/bin/tag.ml
@@ -7,7 +7,7 @@
 open Bos_setup
 open Dune_release
 
-let vcs_tag tag ~dry_run ~commit_ish ~force ~sign ~delete ~msg ~yes =
+let vcs_tag pkg tag ~dry_run ~commit_ish ~force ~sign ~delete ~msg ~yes =
   Vcs.get () >>= fun repo ->
   Vcs.commit_id ~dirty:false ~commit_ish repo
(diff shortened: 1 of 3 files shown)


bin/tag.ml
```diff
@@ -59,9 +59,8 @@ let vcs_tag tag ~dry_run ~commit_ish ~force ~sign ~delete ~msg ~yes =
               commit_ish)
           ~default_answer:Yes)
       >>= fun () ->
-      let msg =
-        match msg with None -> strf "Distribution %s" tag | Some m -> m
```
NathanReb, line 63: We should probably keep this line as well and add the changelog rather than completely replacing it.

NathanReb (commented):
Looks good, thanks! Just have a tiny suggestion and then we can merge!

bin/tag.ml
gpetiot: done

PR #284, opened 2020-09-07
Add a '--token' option to 'publish' and 'opam' commands to specify a github token

### LINKED ISSUE #273: GitHub Token from the environment?
I'm exploring automatic opam releases using GitHub Actions, which can provide a GitHub token through an environment variable. However, it seems that `dune-release` only supports having a token written to the filesystem.

Could this accept a token from the environment before checking for it on the filesystem?

gpetiot:
Fix #273, not sure how to test it yet

### DIFF
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -53,6 +53,18 @@ let opam =
   let docv = "FILE" in
   Arg.(value & opt (some path_arg) None & info [ "opam" ] ~doc ~docv)
 
+let token =
+  let doc =
+    "The github token to use. The regular case is to have a locally store the \
+     github token that $(b,dune-release) can automatically use and to not set \
+     this option. This option is used to override the local configuration \
+     token for example as part of a Github Actions workflow where the github \
+     token is provided through an environment variable."
+  in
+  let docv = "FILE" in
+  let env = Arg.env_var "DUNE_RELEASE_GITHUB_TOKEN" in
+  Arg.(value & opt (some path_arg) None & info [ "token" ] ~doc ~docv ~env)
+
 let keep_v =
   let doc = "Do not drop the initial 'v' in the version string." in
   Arg.(value & flag & info [ "keep-v" ] ~doc)
--- a/lib/delegate.ml
+++ b/lib/delegate.ml
@@ -18,11 +18,12 @@ let run_delegate ~dry_run del args =
 
 (* Publish request *)
 
-let publish_distrib ~dry_run ~msg ~archive ~yes pkg =
+let publish_distrib ?token ~dry_run ~msg ~archive ~yes pkg =
   Pkg.delegate pkg >>= function
   | None ->
       App_log.status (fun l -> l "Publishing to github");
-      Github.publish_distrib ~dry_run ~yes ~msg ~archive pkg >>= fun url ->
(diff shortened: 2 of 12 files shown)


bin/cli.ml
```diff
@@ -53,6 +53,13 @@ let opam =
   let docv = "FILE" in
   Arg.(value & opt (some path_arg) None & info [ "opam" ] ~doc ~docv)
 
+let token =
+  let doc =
+    "The github token to use. If absent looks for a token stored locally."
```
NathanReb, line 58: Maybe we should clarify that it's okay not to pass a token through this option and that's actually the regular usecase. How about mentioning this is used to replace or override the regular config token instead?

bin/cli.ml
```diff
@@ -53,6 +53,13 @@ let opam =
...
+  in
+  let docv = "FILE" in
+  Arg.(value & opt (some path_arg) None & info [ "token" ] ~doc ~docv)
```
NathanReb, line 61: Cmdliner also provides a way to pass it directly through an env variable which I think would also make sense for CI jobs so that the token don't get printed in the CI logs for instance. Could you please add that?

lib/delegate.ml
```diff
@@ -18,11 +18,12 @@ let run_delegate ~dry_run del args =
 
 (* Publish request *)
 
-let publish_distrib ~dry_run ~msg ~archive ~yes pkg =
+let publish_distrib ?token ~dry_run ~msg ~archive ~yes pkg =
```
NathanReb, line 21: How about moving the token selection further up in `bin` and make this arg mandatory. That makes it clear from the beginning where the token comes from.

Same applies to the opam command, I think the selection should be made as early as possible and then the token passed around rather than delaying the filesystem dependent part.

What do you think?

NathanReb: Hmm okay I realize we can't strictly do this because of the delegate case. I guess we'll have to wait until we finally get rid of them!

NathanReb (commented):
I have a few comment but all in all it looks good, thanks for taking care of this!

lib/delegate.ml
gpetiot: That's what I did at first but yes I realized we shouldn't try to determine the token before calling the delegate as it's specific to github.

bin/cli.ml
gpetiot: good idea, done!

bin/cli.ml
gpetiot: done

bin/cli.ml
```diff
@@ -53,6 +53,18 @@ let opam =
...
+let token =
+  let doc =
+    "The github token to use. The regular case is to have a locally store the \
+     github token that $(b,dune-release) can automatically use and to not set \
+     this option. This option is used to override the local configuration \
+     token for example as part of a Github Actions workflow where the github \
+     token is provided through an environment variable."
+  in
+  let docv = "FILE" in
+  let env = Arg.env_var "GITHUB_TOKEN" in
```
NathanReb, line 65: Can you prefix it with `DUNE_RELEASE_` please to make sure we don't accidentally get another token?

gpetiot: done

NathanReb (approved):
Thanks!

PR #290, opened 2020-09-17
Better report GitHub API errors

### LINKED ISSUE #289: Github API errors should be better reported
At the moment we only report the `message` field from the API response but this field alone is not very helpful.
For example, see the following Github API response:
```json
{
  "message": "Validation Failed",
  "errors": [
    {
      "resource": "PullRequest",
      "code": "custom",
      "message": "A pull request already exists for kit-ty-kate:release-opam-0install-0.3."
    }
  ],
  "documentation_url": "https://docs.github.com/rest/reference/pulls#create-a-pull-request"
}
```

Printing the message alone is not giving much feedback on why the request failed, we should probably just print the whole JSON.

gpetiot:
Fix #289

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,7 @@
 - Trim the data of the `url` file before filling the `url.src` field. This fixes an issue that caused the `url.src` field to be a multi-line string instead of single line. (#270, @gpetiot)
 - Fix a bug causing dune-release to exclude all hidden files and folders (starting with `.`) at the
   repository from the distrib archive (#298, @NathanReb)
+- Better report GitHub API errors, all of the error messages reported by the GitHub API are now checked and reported to the user. (#290, @gpetiot)
 
 ### Security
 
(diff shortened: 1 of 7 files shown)


gpetiot:
Can we merge this one?

pitag-ha (commented):
Thanks, I like the PR, it's very well-readable! 

Just one general doubt: in the issue (#290), @NathanReb proposed to log the whole JSON. Here you are logging all `message` values in the JSON, but leaving out the other values. I'm doubting if if might be good to also log (at least some of) the other values for 2 reasons:
-  If I understand the [github documentation](https://developer.github.com/v3/#client-errors) correctly, the objects in `errors` only contain a `message`, if the error is a custom error. If not, the important information seems to be in the error code name (key `code`).
- I think some other values might also be pretty helpful. E.g., the one of `documentation_url` might be interesting to the user.

Why did you decide to only log the `message` values?

gpetiot:
I changed the data representation, from what I get in the github api documentation this kind of handled error should be checked with the deeper messages anyway (for the kind of errors we handle in dune-release for now, that is only when a PR already exists)
I also added the `code` and `documentation_url` fields in the output.

pitag-ha (approved):
Thanks! To me it looks good.

PR #295, opened 2020-10-07
Wrap cmdliner arguments with variants to prevent args order mistakes

NathanReb:
We're using this trick from @CraigFe in duniverse and I thought it'd be a good idea to use it here as well, especially considering the number of args some commands have.

Let me know what you think!


gpetiot (approved):
Looks good!

PR #296, opened 2020-10-07
Simplify distrib tarball preparation

NathanReb:
This used to rely on Distrib.t which was in fact never instantiated to anything than the default value. This simplifies this module to a static list of paths to exclude for now.

Eventually we'll move some `Pkg` functions there but for now I mostly wanted to get rid of all the indirections.

### DIFF
--- a/lib/distrib.mli
+++ b/lib/distrib.mli
@@ -4,38 +4,8 @@
    %%NAME%% %%VERSION%%
   ---------------------------------------------------------------------------*)
 
-(** {1 Distribution description} *)
-
-open Rresult
-
-(** {1:distrib Distribution description} *)
-
-type t
-(** The type for describing distribution creation. *)
-
-val v :
-  ?massage:(unit -> (unit, R.msg) result) ->
-  ?exclude_paths:(unit -> (Fpath.t list, R.msg) result) ->
-  unit ->
-  t
-(** [distrib ~massage ~exclude_paths ()] influences the distribution creation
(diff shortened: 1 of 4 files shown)


lib/distrib.mli
```diff
@@ -4,38 +4,7 @@
...
-    In the following the {e distribution build directory} is a private clone of
-    the package's source repository's [HEAD] when [dune-release distrib] is
-    invoked.
-
-    - [massage] is invoked in the distribution build directory, before
-      archiving. It can be used to generate distribution time build artefacts.
-      Defaults to {!massage}.
-    - [exclude_paths ()] is invoked in the distribution build directory, after
-      massaging, to determine the paths that are excluded from being added to
-      the distribution archive. Defaults to {!exclude_paths}. *)
```
gpetiot, line 34: Maybe these lines of comments can be attached to the `exclude_path` function even if the name is quite explicit.

gpetiot (approved):
Looks good, thanks!

lib/distrib.mli
NathanReb: None of it really applies anymore, we dropped massaging (which never actually did anything) and the current documentation seems to speak for itself. What would you addd to make it more complete?

NathanReb: Hmm it seems I messed up and didn't commit the doc, let me fix that!

gpetiot: Oh I was just referring to "the paths that are excluded from being added to the distribution archive"

PR #298, opened 2020-10-08
Fix a bug excluding all root dotfiles from the generated distrib tarball

NathanReb:
`Bos.OS.Dir.fold_contents` behaviour though following its spec is very counter intuitive since specifying `~dotfiles:true` will only include dotfiles and folders in subdirectories.

Luckily this hasn't broken anything yet but it is obviously a bad behaviour that needed fixing. The fact that we maintain an explicit
list of files and folders to exclude from the archive, some of which are root dotfiles and folders confirms that this behaviour was not intentional.


NathanReb:
Hmm wait this is probably worth a test, let me add that!

NathanReb:
And there you go!

gpetiot (approved):
Looks good to me!

NathanReb:
Thanks, I'll merge since both travis and ocaml-ci lint check are green, I think the remaining ocaml-ci jobs are awaiting for workers :/

PR #299, opened 2020-10-08
Include versioned files to the archive

NathanReb:
Those special rules resulted in removing versioned files in the archive. There's no need for special treatment of build dirs since those shouldn't be versioned anyway and therefore won't be included in the archive since we go through a clone and a checkout before building the tarball.


gpetiot (approved):
It shouldn't be necessary indeed, should we add a changelog entry anyway?

NathanReb:
Yep, I just pushed this in a rush yesterday and didn't take time to write it. I knew the check would catch it and it wouldn't be merged accidentally, I really love this check!

PR #300, opened 2020-10-09
Submodules in tarball

### LINKED ISSUE #279: Support Submodules
Submodules are currently ignored when dune-release creates an archive. It would be much better if it was possible to include them as part of the source.

Perhaps there should be an option to switch off this behavior. I know some use submodules for brining into scope normal dependencies during development.

NathanReb:
Fixes #279 

This PR add a new feature allowing users to include their git submodules to the distrib tarball if they use them for vendoring dependencies.

This is enabled by passing the `--include-submodules` to the `dune-release distrib` command. The flag can also be passed to the
`dune-release` or `dune-release bistro` command with the same effect.

When the flag is passed, the distribution creation tarball will clone the repo and checkout to the right commit as usual. At this point it will run `git submodule update --init` from the checked out local clone folder, resulting in a pull of the submodules to the right commit (i.e. as defined in `.gitmodules`). The regular tarball creation process will resume from their and will include the new sources to the tarball, with the exception of the `.git` file that we explicitly ignore.

Regarding the `.git` file, I went for the easy solution there, by ignoring all `.git` files and folders accross the repository as I don't think it can be harmful to do so. If I'm proven wrong we can always refine this to ignore only the root `.git` folder and any `.git` files in the root of all registered submodules.

For now the feature is quite minimal and I thought of a few follow ups:
- pass the `--recursive` flag to `git submodule update` so that submodules are recursively pulled. This might be required in some cases.
- Allow passing a list of path to `--include-submodules` to restrict which submodules should be included.

For now I think we should settle for the simplest implementation and await feedback before making this more complicated than it needs to be, although none of the above mentioned features would be too hard to implement.


NathanReb:
CC @rgrinberg, could you give this a spin with ocaml-lsp and see if it fits your needs? That's where I tested it myself so hopefully it should work fine but I don't know the specifics of the project so getting your approval on this would be great!

lib/vcs.ml
```diff
@@ -146,6 +146,9 @@ let git_tag ~dry_run r ~force ~sign ~msg ~commit_ish tag =
 let git_delete_tag ~dry_run r tag =
   run_git_quiet ~dry_run r Cmd.(v "tag" % "-d" % tag)
 
+let git_submodule_update ~dry_run r =
+  run_git_quiet ~dry_run r Cmd.(v "submodule" % "update" % "--init")
```
kit-ty-kate, line 150: Should it also use `--recursive` like done in `ocaml-ci`/`ocurrent`?

See https://github.com/ocurrent/ocurrent/blob/a8909d5669bcb2428ccdc27591011c1e5897dbc3/plugins/git/cmd.ml#L48

kit-ty-kate: Ah sorry I should've read the original description of this PR, you already suggested it ^^"

NathanReb: As I mentioned in the PR message, it might make sense to do so but I'd be curious to get some feedback before making this the default. I think if we start relying to heavily on submodules that way, we could run into problems such as:
I have two submodules a and b and both of them have a the same c submodule. If we pull recursively c will be duplicated and assuming a, b and c use dune, this might fail.

If we don't recursively pull submodules, the use can chose to include c himself.

I feel like it is easier to work around those issues if `--recursive` isn't the default or is at least configurable. Making it configurable might prove a bit of a headache from a user interface perspective and I'd rather not dive into it before someone actually expresses the need for it!

Do you actually need the `--recursive` in one of your project?

kit-ty-kate: > Do you actually need the `--recursive` in one of your project?

No it was just a small remark, no worries. It worked great for me so far, I'll post a proper comment once CI is green on the PR I've made using it.

kit-ty-kate:
Tested in https://github.com/ocaml/opam-repository/pull/17383 successfully. CI did not see any issues and the package was thus merged a few hours ago.

aryx:
Can we merge this ASAP so @rgrinberg can release ocaml-lsp in OPAM and other people can use ocaml-lsp in their own project?

gpetiot (approved):
Looks good to me, thanks!

NathanReb:
I'm waiting for Rudi's feedback on this, I want to make sure it fits the ocaml-lsp usecase before merging. He confirmed he will try to release ocaml-lsp with this branch shortly!

lib/pkg.ml
```diff
@@ -515,21 +515,30 @@ let assert_tag_exists ~dry_run repo tag =
   if Vcs.tag_exists ~dry_run repo tag then Ok ()
   else R.error_msgf "%s is not a valid tag" tag
 
-let distrib_archive ~dry_run ~keep_dir p =
+let pull_submodules ~dry_run ~dist_build_dir =
+  Sos.with_dir ~dry_run dist_build_dir
+    (fun () -> Vcs.get () >>= Vcs.submodule_update ~dry_run)
+    ()
+  |> R.join
```
gpetiot, line 522: as it doesn't involve a `Pkg.t` type wouldn't it better to put this function in vcs.ml ?

rgrinberg:
> I'm waiting for Rudi's feedback on this, I want to make sure it fits the ocaml-lsp usecase before merging. He confirmed he will try to release ocaml-lsp with this branch shortly!

I've submitted lsp to opam and everything appears to work fine. Thanks @NathanReb 

@aryx your enthusiasm is much appreciated as well ;)

NathanReb:
I just rebased so we should merge once CI goes green!

PR #301, opened 2020-10-12
Add more ellipses to the `no-doc` cram test

pitag-ha:
It seems like the `no-doc`-test isn't fully deterministic. This commit
adds a couple of ellipses that are necessary for it to pass on some
environments.


gpetiot (approved):
Looks good!

PR #302, opened 2020-10-16
Improve the deprecation of delegates

pitag-ha:
This PR improves the deprecation of delegates as suggested by @NathanReb in the review of #276: the deprecation warnings in the docs now come after the corresponding concept has been introduced; at important places, the new workflow gets explained after the deprecation warning; functions that were moved to the deprecation module have been moved back to where they were; the deprecation module now has a more general name in order to also deprecate other things there.


NathanReb:
Any idea where the CI failures are coming from?

NathanReb (commented):
Looks good! Can you please add an `.mli` file (with a bit of documentation eventually) for `deprecate.ml` though?

Also I think we need to add warnings in `Delegate.publish_doc` and `Delegate.publish_distrib` when we go through the delegate path.

pitag-ha:
I've added a `deprecate.mli` file. 

I've also added a warning for `publish_distrib`. I think the warning for `publish_doc` is [here](https://github.com/ocamllabs/dune-release/pull/302/files#diff-77ca13fd2aa0b2234796f027396bf6b2db6fbaaae716857cf3d9885a025a12d9R46).

NathanReb (approved):
Awesome possum!

PR #303, opened 2020-10-24
Set the default tag message to "Release <tag>" instead of "Distribution <tag>"

kit-ty-kate:
When making a release, the default title dune-release will give it when calling `dune-release tag` is `Distribution <tag>`.
This title will be carried on in the github UI:
scrn-2020-10-24-02-22-46
However I feel like this is a bit surprising. I feel like "distribution" doesn't quite convey the right meaning and "release" feels more right in the default case.


kit-ty-kate:
Any change anyone could have a look at this?

gpetiot (approved):

NathanReb:
I think this was partially fixed by https://github.com/ocamllabs/dune-release/pull/338. Probably not a bad idea to change the tag message though!

NathanReb (approved):

PR #304, opened 2020-10-26
Suggest a solution for users without ssh set up

### LINKED ISSUE #212: dune-release fails on default config of macos

pitag-ha:
Users who haven't set up ssh will get a Permission denied error from
github when using `dune-release publish`. This commit adds to the README
how the user can configure git in order to solve that problem. It also
adds a hint to the Permission denied error message pointing to that
new part of the README.

Closes #212


NathanReb (approved):
Looks good to me, thanks!

NathanReb:
We need to fix the CI!

gpetiot (approved):
Looks good, thanks!

PR #305, opened 2020-10-27
Fix CI

NathanReb:


tests/bin/no_doc/run.t
```diff
@@ -40,206 +40,17 @@ We need to set up a git project for dune-release to work properly
...
-
-    $ dune-release publish doc
+    $ dune-release publish doc -y --dry-run
     [-] Publishing documentation
-    [-] Selected packages: whatever whatever-lib
-    [-] Generating documentation from _build/whatever-0.1.0.tbz
-    [-] Publishing to github
-    dune-release: [ERROR] Could not derive publication directory $PATH from opam doc field value ""; expected the pattern $SCHEME://$USER.github.io/$REPO/$PATH
+    ...
+    dune-release: [ERROR] directory contents _build/whatever-0.1.0: No such file or directory
```
gpetiot, line 48: I think it's important to test the previous message that is only printed in non dry-run mode and still required to call dune-release distrib before, something like

```
    $ dune-release publish doc
    [-] Publishing documentation
    [-] Selected packages: whatever whatever-lib
    [-] Generating documentation from _build/whatever-0.1.0.tbz
    [-] Publishing to github
    dune-release: [ERROR] Could not derive publication directory $PATH from opam doc field value ""; expected the pattern $SCHEME://$USER.github.io/$REPO/$PATH
    [3]
```

I agree on the end of the diff though

NathanReb: I'm not sure it is relevant to test the doc publication skip. I think this test is there to ensure that by default, doc publication is skipped if the field is missing but is still run if explicitly asked for, not to test the behavior of doc publication in general.

The more output we leave in those test the more likely they are to break for unrelated changes and that's something we want to avoid, especially with the ellipsis which makes it hard to use promotion.

If we want to test which packages are selected, or how the doc is generated I believe that it should be a separate test and should probably rely on the resulting doc folder rather than on the logs.

These cram tests definitely are not perfect and it's hard to test dune-release without network access and a hell of a setup but that's precisely why I think we should keep them as simple as possible. Here we're basically just testing two branches of an if. At some point we might want to re-architecture this feature so it can be unit tested but for now I'm happy with a "simple" cram test.

Is that the reason why you're suggesting we keep a wider part of the output or do you have something else in mind?

gpetiot: > Is that the reason why you're suggesting we keep a wider part of the output or do you have something else in mind?

No it was only for this, but alright I agree then, we can merge as is!

gpetiot (approved):

PR #318, opened 2020-12-08
Add `check` command

pitag-ha:
The PR adds the new `check` command which allows the user to make sure that the whole release process will be smooth: it checks if the opam-file is dune-release compatible and - unless stated otherwise by the user via the skip options - performs the checks that are done on the distribution archive: if it builds, if it passes its tests and if it passes the dune-release lints. 

For all that, the same options are taken into account as will be when releasing. I've left out the options `--dist-opam`, `--readme` and `--change-log`, since they are also left out in `dune-release lint`. But it would probably be good to make that consistent at some point..

Before the commit adding the `check` command there are a couple of commits improving and fixing some minor issues in that context (see commit messages).


bin/check.ml
```diff
@@ -0,0 +1,70 @@
...
+    (`Working_tree on_working_tree) =
+  ( if on_working_tree then
+    OS.Dir.current () >>= fun dir ->
+    Check.check_project ~name ~pkg_names ~tag ~skip_lint ~skip_build ~skip_tests
+      ~dir
+  else
+    Vcs.get_tag tag version >>= fun inferred_tag ->
+    Vcs.get () >>= fun repo ->
+    assert_tag_exists repo inferred_tag >>= fun () ->
+    Fpath.of_string "_build/.dune-release-check" >>= fun dir ->
```
gpetiot, line 27: Shouldn't _build be parameterized with `--build-dir` ?

bin/check.ml
```diff
@@ -0,0 +1,70 @@
...
+
+let doc = "Check dune-release compatibility"
+
+let man =
+  [
+    `S Manpage.s_description;
+    `P
+      "$(tname) checks if the whole release process can be done with \
+       dune-release, assuming that you'll provide the same options as \
+       providing here. With the [--working_tree] option, you can perform the \
```
gpetiot, line 60: I would say "as provided here"
also typo on the option name, should be `--working-tree`

lib/check.mli
```diff
@@ -0,0 +1,33 @@
...
+  name:string option ->
+  pkg_names:string list ->
+  skip_lint:bool ->
+  skip_build:bool ->
+  skip_tests:bool ->
+  tag:string option ->
+  dir:Fpath.t ->
+  (int, [ `Msg of string ]) result
+
+(* fixme*)
```
gpetiot, line 23: is this fixme still necessary?

lib/check.mli
```diff
@@ -0,0 +1,33 @@
...
+  dry_run:bool ->
+  skip_build:bool ->
+  skip_tests:bool ->
+  pkg_names:string list ->
+  log_status:bool ->
+  Fpath.t ->
+  (int, [ `Msg of string ]) result
+
+(* checks if the packages in [pkg_names] - or, if [pkg_names] is empty, all packages in [dir] - build and pass their tests; depending on [log_status] with or without status logs. 
+  if [skip_build] or [skip_tests], the correspondent checks are skipped. *)
```
gpetiot, line 11: Why is this not a documentation comment?

lib/pkg.ml
```diff
@@ -233,48 +226,65 @@ let path_of_distrib p =
...
+    (if a || b then "releases/download/$(TAG)/" else "releases/") ^ basename
+  in
+  name p >>= fun name ->
+  tag p >>= fun tag ->
+  let defs = String.Map.(empty |> add "NAME" name |> add "TAG" tag) in
+  Pat.of_string uri >>| Pat.format defs
+
+let uri_of_dev_repo p =
+  opam_field_hd p "dev-repo"
+  >>| Option.map (fun dev_repo ->
```
gpetiot, line 240: for 4.06 and 4.07 we must use `Stdext.Option`

gpetiot (commented):
Besides a few minor comments it looks good to me, thanks! Don't forget to add a changelog entry though

lib/vcs.mli
```diff
@@ -68,6 +68,9 @@ val tag_exists : dry_run:bool -> t -> string -> bool
 
 val tag_points_to : t -> tag:string -> string option
 
+val get_tag :
+  string option -> string option -> (string, [ `Msg of string ]) result
```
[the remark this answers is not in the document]
pitag-ha: Yes, that's a good idea. I've done that now, thanks!

bin/check.ml
pitag-ha: It totally should. Done. Thanks!

bin/check.ml
pitag-ha: > I would say "as provided here"

I've just skipped the word "providing"/"provided" now. I think "as here" alone is clear enough. Does that sound good to you?

> also typo on the option name, should be --working-tree

good catch!

lib/check.mli
pitag-ha: Now not anymore. I've removed it

lib/check.mli
pitag-ha: Good catch!

lib/pkg.ml
pitag-ha: Thanks! I didn't know about that.

bin/check.ml
gpetiot: Yes it's better, thanks!

lib/pkg.ml
```diff
@@ -561,20 +576,21 @@ type f =
   dir:Fpath.t ->
   args:Cmd.t ->
   out:(OS.Cmd.run_out -> (string * OS.Cmd.run_status, Sos.error) result) ->
+  ?err:Bos.OS.Cmd.run_err ->
```
pitag-ha, line 579: I wasn't sure here if to ignore the dune error output both in `check` and when building/testing in `distrib` or just in the first. Actually I think in both would be better, since the dune message pollutes the dune-release message a lot and `dune build`/`dune runtest` can be run separately if any of them fails. But I've opted for only in `check` to not change the output of `distrib` without necessity. What do you think?

bin/help.ml
```diff
@@ -45,10 +45,7 @@ let release =
         \        your source repository lints and that it builds in the \
          current build\n\
         \        environment and that the package tests pass.";
-      `Pre
-        "dune-release lint\n\
-         dune build # Check out the generated opam install file too\n\
```
pitag-ha, line 50: What does "Check out the generated opam install file too" mean? Should I keep that there?

tests/bin/check/run.t
```diff
@@ -0,0 +1,130 @@
...
+If the condition described above is fulfilled, there are 5 checks to be performed
+
+    $ cat > my_pkg.opam <<EOF \
+    > opam-version: "2.0"\
+    > dev-repo: "git+https://github.com/fu/fa.git"\
+    > description : "Some description"\
+    > EOF
+    $ echo "(lang dune 2.7)" > dune-project
+
+    $ dune-release check --dry-run --working-tree | sed "s/lint of .* and/lint of <project_dir> and/" | sed "s/.*] package(s) pass the tests/[<status>] package(s) pass the tests/" | sed '/chdir/d' | sed '/=>/d' | sed '/-: exec/d'
```
pitag-ha, line 23: I've tried to put these `sed` calls that get repeated several times into functions, but got an sh error. Is it possible to do that in mdx cram tests?

gpetiot: I think it's simpler to define a function in a separate sh file, like I did in #333 for sanitize.sh

pitag-ha: That's a good idea! Also using dune cram tests instead of mdx cram tests seems great. I'll wait for that PR to be merged to use dune cram tests here then. Or is it possible to do what you're saying with mdx?

gpetiot: I think it should be possible with mdx as well

NathanReb: It might also be possible to mix mdx and dune cram test so we don't have to migrate as much test when doing the move. I.e. new tests should be written as dune cram tests.

Also if you need to sanitize the output, I'd be in favor of an OCaml script rather than a bash one for portablitiy of the tests and ease of maintenance!

bin/help.ml
gpetiot: I think this refers to `dune-release distrib` that will run the lint on the archive, then checks that it builds, then checks the tests, it's not the doc of `dune-release lint`

tests/bin/check/run.t
gpetiot: If we can make dune cram tests cohabitate with the current tests that would be better indeed

lib/pkg.ml
```diff
@@ -431,21 +440,27 @@ let infer_from_readme dir =
...
           infer_from_readme dir >>= function
-          | Some n -> Ok n
-          | None ->
-              Logs.err (fun m ->
-                  m
-                    "cannot determine distribution name automatically: add \
-                     (name <name>) to dune-project");
-              exit 1))
+          | Some n -> Ok (Some n)
+          | None -> Ok None))
```
gpetiot, line 452: What is the value of returning `Ok None` ? I think we can rewrite this function to return `(string, ['Msg of string]) result` so we can return the message directly instead of having `infer_name_err` on the side, and probably the èinfer_name` shouldn't return `exit 1` but let propagate the error value up up up?

lib/pkg.ml
```diff
@@ -74,11 +74,7 @@ let name p = Ok p.name
 
 let with_name p name = { p with name }
 
-let tag p =
-  match (p.tag, p.version) with
-  | Some v, _ -> Ok v
-  | None, None -> Vcs.get () >>= fun r -> Vcs.describe ~dirty:false r
-  | None, Some v -> Ok v
+let tag p = Vcs.get_tag ?tag:p.tag ?version:p.version ()
```
NathanReb, line 77: Is there any particular reason why you moved this logic from this module to `Vcs` ?

I think that `Vcs` should stick to running VCS commands and converting their output into OCaml types, this logic is specific to dune-release from the look of it.

NathanReb (commented):
This looks good but I think there are a few things to improve and clarify. I left a few review comments but I also have things to report here.

I tried it locally on a dummy project and got the following:
```
[ OK ] main package dummy.opam is dune-release compatible.
[ OK ] connected to internet
[ OK ] package(s) build
[ OK ] package(s) pass the tests
[FAIL] lint of _build/.dune-release-check and package dummy failure: 1 errors.
For information on the lint errors, run `dune-release lint`.
```
I think the lint report should be printed here, having to run a separate command to get the results seems kind of unfriendly to the user. Is there a particular reason that made it hard to show the lint results here?

Another thing that strike me is the part about the internet connection. I think that if we want to go down that road there are better things to check like does the repo exist, does the user opam fork exists, is the token properly set and does the user can talk to the github API, do they have their ssh key properly set etc...
In a first iteration I think it is safer to stick to strictly local checks, we can always think of how to improve this further later. The current internet check probably doesn't add much value.

Also a quick note: I think the PR needs a rebase!

bin/help.ml
pitag-ha: Isn't this the help page of `dune-release help release`?

tests/bin/check/run.t
pitag-ha: Thanks to both for your advice! I've separated the functions into separate files now and have written them in OCaml.

lib/pkg.ml
pitag-ha: > What is the value of returning Ok None ?

The return type of this function gets propagated from the return type of the `infer_from...` functions. So the value of `Ok None` is the same as in the `infer_from...` functions:
there was no sys call error -> `Ok`;
there's not enough information in any of the resources to infer the name from them -> `None`

> and probably the `infer_name` shouldn't return `exit 1`

I was also wondering about that behaviour here. I've just preserved the behaviour. Since I didn't know why it was done that way in the first place, I didn't feel confident changing it. Do you happen to know why it was done like that?

CHANGES.md
```diff
@@ -61,6 +62,7 @@
 - Fixes a bug under cygwin where dune-release was unable to find the commit hash corresponding to the release tag (#329, @gpetiot)
 - Fixes release names by explicitly setting it to match the released version (#338, @NathanReb)
 - Fix a bug that prevented release of a package whose version number contains invalid characters for a git branch. The git branch names are now sanitized. (#271, @gpetiot)
+- `publish`: Fix the process of inferring user name and repo from the dev repo uri (#318, @pitag-ha)
```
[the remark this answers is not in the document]
pitag-ha: Done. Since I need that commit for the `check` command to work, I've still left the commit in this PR for now and converted this PR to draft. I'll fix that and rebase once [the extracted PR](https://github.com/ocamllabs/dune-release/pull/348) is merged.

bin/check.ml
```diff
@@ -0,0 +1,77 @@
...
+  Vcs.clone ~dry_run:false ~force:true repo ~dir >>= fun () ->
+  Vcs.get ~dir () >>= fun clone_vcs ->
+  Vcs.checkout ~dry_run:false clone_vcs ~branch:"dune-release-check"
+    ~commit_ish:tag
+
+let check (`Dry_run dry_run) (`Package_names pkg_names)
+    (`Package_version version) (`Dist_tag tag) (`Keep_v keep_v)
+    (`Build_dir build_dir) (`Skip_lint skip_lint) (`Skip_build skip_build)
+    (`Skip_tests skip_tests) (`Working_tree on_working_tree) =
+  (if on_working_tree then
```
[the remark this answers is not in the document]
pitag-ha: If I did that I'd need to do another `if` at the end to know if I need to delete the clone directory or not. I feel like that would be a bit confusing. What do you think?

bin/check.ml
```diff
@@ -0,0 +1,77 @@
...
+    | None -> Fpath.of_string "_build")
+    >>= fun build_directory ->
+    let dir = Fpath.(build_directory // v ".dune-release-check") in
+    clone_and_checkout_tag repo ~dir ~tag:inferred_tag >>= fun () ->
+    let check_result =
+      Check.check_project ~dry_run ~pkg_names ?tag ?version ~keep_v ?build_dir
+        ~skip_lint ~skip_build ~skip_tests ~dir ()
+    in
+    let () =
+      match Sos.delete_dir ~dry_run:false ~force:true dir with
```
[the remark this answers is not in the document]
pitag-ha: Good catch, thanks!

> To be honest I don't think check should allow dry-run

I think the same, but `lint` allows dry-run and I wanted to be coherent. I guess we're going to remove dry-run from `lint` in 2.0.0. Is that right? Thinking about dry-run in `check` twice, I also think it's better to directly not have it (even being incoherent with `lint` for a short time). Is that what you meant? If so, I change it.

Btw, there was a similar problem with some of the command line options: `--dist-opam`, `--readme` and `--change-log`. I think both `lint` and `check` should have them. But `lint` doesn't at the moment. I decided to be coherent with `lint` there, too. What do you think about that?

lib/pkg.ml
pitag-ha: I've moved the inner logic of the function (but left the function itself), since I need to get the tag before having created a package. But I hadn't noticed the `Vcs` pattern of always running commands and converting their output. Where do you think that function might fit into?

pitag-ha:
> Is there a particular reason that made it hard to show the lint results here?

Letting the lint results get logged directly would look as follows:
```
[ OK ] main package dummy.opam is dummy compatible.
[ OK ] package(s) build
[ OK ] package(s) pass the tests
[ OK ] File README is present.
[FAIL] File LICENSE is missing.
[ OK ] File CHANGES is present.
[ OK ] File opam is present.
[ OK ] lint opam file dummy.opam.
[ OK ] opam field description is present
[ OK ] opam fields homepage and dev-repo can be parsed by dummy
[ OK ] opam field doc can be parsed by dummy
[FAIL] lint of _build/.dummy-check and package dummy failure: 1 errors.
```

In a 2-package project it would even look like this:
```
[ OK ] main package dummy.opam is dummy compatible.
[ OK ] package(s) build
[FAIL] package(s) pass the tests
[ OK ] File README is present.
[FAIL] File LICENSE is missing.
[ OK ] File CHANGES is present.
[ OK ] File opam is present.
[ OK ] lint opam file dummy_sub.opam.
[ OK ] opam field description is present
[ OK ] opam fields homepage and dev-repo can be parsed by dummy
[ OK ] opam field doc can be parsed by dummy
[FAIL] lint of /home/sonja/Documents/Tarides/projects/dummy and package dummy_sub failure: 1 errors.
[ OK ] File README is present.
[FAIL] File LICENSE is missing.
[ OK ] File CHANGES is present.
[ OK ] File opam is present.
[ OK ] lint opam file dummy.opam.
[ OK ] opam field description is present
[ OK ] opam fields homepage and dev-repo can be parsed by dummy
[ OK ] opam field doc can be parsed by dummy
[FAIL] lint of /home/sonja/Documents/Tarides/projects/dummy and package dummy failure: 1 errors.
```

So I think that the lint output would pollute the output quite a bit. But you're right that having to run a separate command if the lint fails isn't a great solution either. I figured that for now it didn't matter to much, since in 2.0.0 we should improve the lint anyways (in multi-package projects, only run the hygiene checks once since they are package independent). 

What do you think about that output "pollution"? Do you think that would be ok for now (and better than having to run another commend)? Maybe I could figure out if it would be easy to indent the lint output (to make clear which check comes from where).

> I think that if we want to go down that road there are better things to check like does the repo exist, does the user opam fork exists, is the token properly set and does the user can talk to the github API, do they have their ssh key properly set etc...

That's a very good point! I've removed the internet check and left it on a different branch in case at some point we want to check more things.

bin/help.ml
gpetiot: Oh yes right, well it looks like it's referring to the `<pkgname>.install` file that can be generated by dune build (didn't find any locally), but I think we can remove this line indeed!

lib/pkg.ml
pitag-ha: Ok, done.

bin/check.ml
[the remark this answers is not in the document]
pitag-ha: Yes, that's a nice idea. Thanks!

lib/pkg.ml
pitag-ha: I've moved it to `Pkg` now (so now there are two functions to get the tag in `Pkg`).

CHANGES.md
```diff
@@ -11,6 +11,7 @@
   `--include-submodules` flag to `dune-release`, `dune-release bistro` or
   `dune-release distrib` (#300, @NathanReb)
 - Support 'git://' scheme for dev-repo uri (#331, @gpetiot)
+- Add a new command `check` to make sure the release will be smooth (#318, @pitag-ha)
```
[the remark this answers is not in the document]
pitag-ha: Yes, that makes it clearer. Thanks!

NathanReb:
Can you rebase it please?

gpetiot (approved):

tests/bin/check/run.t
```diff
@@ -0,0 +1,194 @@
...
+    [FAIL] cannot determine distribution name automatically: add (name <name>) to dune-project
+    [1]
+
+Let's clean up to not interfere with the following tests
+
+    $ rm first_pkg.opam other_pkg.opam
+
+If the condition described above is fulfilled, there are 4 checks to be performed
+
+    $ cat > my_pkg.opam <<EOF \
```
NathanReb, line 16: Wouldn't it make more sense to move this part right after the `rm` line above?

tests/bin/check/run.t
```diff
@@ -0,0 +1,194 @@
...
+    > description : "Some description"\
+    > maintainer : "me"\
+    > authors : "also me"\
+    > homepage : "https://github.com/fu/fa"\
+    > bug-reports : "https://github.com/fu/fa/issues"\
+    > synopsis : "Dope project"\
+    > EOF
+    $ echo "(lang dune 2.7)" > dune-project
+
+    $ dune-release check --working-tree | ./make_lint_directory_deterministic.exe | ./make_test_directory_deterministic.exe
```
NathanReb, line 28: Might be good to merge those two binaries into a single one, what do you think?

NathanReb: We could even move it upwards in `tests/bin/helpers/sanitize_output.ml` so it can be enriched and reused accross our different cram tests!

What do you think about that?

NathanReb (commented):
Looks good!

I just have minor comments regarding the tests but otherwise it's all :+1:

tests/bin/check/run.t
pitag-ha: I've done both things. Thanks for the help!

tests/bin/check/run.t
pitag-ha: Ok, I've separated each test set up from its test in general now.

tests/bin/helpers/make_check_deterministic.ml
```diff
@@ -0,0 +1,22 @@
+open Re
+
+let make_lint_directory_deterministic line =
+  (* OCaml implementation of bash's
+     sed "s/lint of .* and/lint of <project_dir> and/" *)
+  let re = compile @@ seq [ str "lint of "; rep any; str " and" ] in
+  replace_string re ~by:"lint of <project_dir> and" line
+
+let makte_test_directory_deterministic line =
```
gpetiot, line 9: nitpick: slight typo in `make`

pitag-ha: Thanks. It's fixed now.

NathanReb (approved):

PR #327, opened 2020-12-11
Remove option --name

### LINKED ISSUE #324: dune-release distrib --name does not suffice to set name
`dune-release distrib` calls `dune subst`, which requires a `(name <name>)` stanza in dune-project. Therefore, trying to set the name with `dune-release distrib --name <name>` is bound to fail with errors like:

~~~
+    Error: The project name is not defined, please add a (name <name>) field to
+    your dune-project file.
+    dune-release: [ERROR] run ['dune' 'subst']: exited with 1
+    [3]
~~~

One solution might be to remove `--name` from `dune distrib`, and mandate a `(name <name>)` stanza in dune-project. Another one would be to patch `dune subst` to accept an argument to set the project name.

lehy:
This removes option --name from all commands.

This fixes #324, where `dune-release distrib` would fail if a name is not present in `dune-project` (because it calls `dune subst`), therefore rendering pointless any passed --name.

So this pull request takes the brute force approach and removes --name from all commands. The upside is that this is simple, consistent and mostly removes code.

The downsides are:

- it causes a possible regression if people were using --name from other commands than `distrib` (I have no idea if that was useful or actually used ; I believe that for people with only one .opam file there would not have been a reason to use this)

- it removes the documentation for how the distribution is named (longest common prefix of .opam files, first word of main README title) : I'm not sure where to put it or even if it matters at all

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,11 @@
 
 ### Removed
 
+- Option --name is removed from all commands. When used with
+  `dune-release distrib`, it was previously effectively ignored. The
+  alternative is to add a `(name <name>)` stanza to
+  `dune-project`. (#324, @lehy)
+
 ### Fixed
 
 - Fix the priority of the `--distrib-uri` option in `dune-release opam pkg`.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,9 +29,9 @@
 ### Removed
 
 - Option --name is removed from all commands. When used with
-  `dune-release distrib`, it was previously effectively ignored. The
-  alternative is to add a `(name <name>)` stanza to
-  `dune-project`. (#324, @lehy)
+  `dune-release distrib`, it was previously effectively ignored. Now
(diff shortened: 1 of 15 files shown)


CHANGES.md
```diff
@@ -28,6 +28,11 @@
 
 ### Removed
 
+- Option --name is removed from all commands. When used with
+  `dune-release distrib`, it was previously effectively ignored. The
+  alternative is to add a `(name <name>)` stanza to
+  `dune-project`. (#324, @lehy)
```
gpetiot, line 34: This should be #327 instead.
Personally I would replace "The alternative is to add ..." with "Now it is required to add ..."

gpetiot (approved):
It looks good to me, thanks! Just some minor comments on the changelog.

CHANGES.md
lehy: Done.

PR #332, opened 2020-12-21
Make the automatic dune-release workflow to stop if a step exits with a non-zero code

gpetiot:
I think it's better to return the intermediate non-zero result as soon as it happens so we can have a more precise log, instead of just carrying on and ignoring the returned exit code of each step except the last one like it's currently the case, what do you think?

Actually some of dune-release steps could return unit instead of a return code but we can keep it like this for now in case we need these exit codes in the future. But personally I would replace these by unit, and use a sum/variant type instead.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@
   This change allows users to push the tag manually to avoid using that code. (#219, @Julow)
 - Recursively exclude all `.git`/`.hg` files and folders from the distrib
   tarball (#300, @NathanReb)
+- Make the automatic dune-release workflow to stop if a step exits with a non-zero code (#332, @gpetiot)
 
 ### Deprecated
 
--- a/bin/bistro.ml
+++ b/bin/bistro.ml
@@ -6,6 +6,9 @@
 
 open Bos_setup.R.Infix
 
+(* Only carry on when the first operation returns 0 *)
+let ( >! ) x y = match x with Ok 0 -> y | _ -> x
+
 let bistro () (`Dry_run dry_run) (`Package_names pkg_names)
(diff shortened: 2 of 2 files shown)


NathanReb (approved):
I definitely agree here, to be honest I think it's a mistake that it returns ok with a non-zero exit status, if anything it should be a ``(unit, (int * [>`Msg of string])) result``, `Ok ()` should imply zero exit status.

I also agree that this is a good first step and we can merge as is, improving the whole thing later.

Out of curiosity, do you know of situations where we explicitly return `Ok x` with `x != 0` ?

CHANGES.md
```diff
@@ -23,6 +23,7 @@
   This change allows users to push the tag manually to avoid using that code. (#219, @Julow)
 - Recursively exclude all `.git`/`.hg` files and folders from the distrib
   tarball (#300, @NathanReb)
+- Make the automatic dune-release workflow to stop if a step doesn't return 0 (#332, @gpetiot)
```
NathanReb, line 26: How about clarifying that with `to stop if a step exits with a non-zero code`? I think the `return 0` part might be a bit confusing whereas a slightly different phrasing makes it clear that it means one of the individual commands failed.

What do you think?

gpetiot:
> Out of curiosity, do you know of situations where we explicitly return Ok x with x != 0 ?

Yes for example `dune-release distrib` returns 1 if the lint or build or tests failed, and `dune-release opam` also returns 1 in a few cases.

CHANGES.md
gpetiot: I agree

PR #333, opened 2021-01-06
Use dune 2.7 cram feature instead of mdx for cram tests

gpetiot:
Not critical but it would be nice to drop the dependency


Leonidas-from-XIV:
Obsoleted by #400 therefore closing this.

PR #334, opened 2021-01-12
Make git related tests more robust

sternenseemann:
* ~~When running git init we pass -b main to set an initial branch
  explicitly. This prevents the hint informing you that git is
  defaulting to master from being printed when using more recent git
  versions.~~
* When running git init we also redirect stderr to /dev/null as
  newer versions of git print a hint to stderr if there's no default
  main branch name set and it is defaulting to 'master'. An alternative
  would be to pass -b, but that is not supported by older git versions
  as seen on debian, for example.
* after git init we set user.name and user.email for the repository, so
  the tests don't fail if they are executed in an environment without a
  global git config (CI, sandboxed build systems, …)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,7 @@
 - Recursively exclude all `.git`/`.hg` files and folders from the distrib
   tarball (#300, @NathanReb)
 - Make the automatic dune-release workflow to stop if a step exits with a non-zero code (#332, @gpetiot)
+- Make git-related mdx tests more robust in unusual environments (#334, @sternenseemann)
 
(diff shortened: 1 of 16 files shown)


sternenseemann:
Alright, seemingly `-b` is a new flag, will have to look for an alternative.

sternenseemann:
Okay redirecting stderr works as an alternative (checked locally against a newer git version).

gpetiot (approved):
Great, thanks! This fixes the CI that was broken since the addition of alpine in the build list. I rebased and added the same changes to the last cram test that was added in the meantime.

PR #338, opened 2021-01-26
Fix release names

### LINKED ISSUE #337: Release aren't named after the version anymore
Recent changes appear to have broken the release creation as they no longer are named after the version but after the tag message.

We need to add a test that we explicitly set the release name when creating it.

NathanReb:
Fixes #337 

We used to rely on github's default name but it appears it changed recently. Setting this explicitly seems like the right thing to do.

@gpetiot you'll probably have to rebase #271 on top of this, in particular `tag_name` and `name` will obviously have different values, `name` being the actual version while `tag_name`will need to be sanitized in the same way the tag is!


NathanReb:
I also slightly refactored `Curl` to use Yojson rather than creating json strings ourselves!

NathanReb:
I need to test this. I'll update the changelog once I did!

gpetiot (approved):
Looks good to me!

PR #343, opened 2021-02-09
Upgrade to opam 2.1

kit-ty-kate:
opam 2.1 hasn't been released yes so I'll keep that as a draft for now.
I've been using this branch for a few months now and haven't noticed any issues yet.


NathanReb:
This looks good but we'll need https://github.com/ocaml/opam-file-format/pull/32 to preserve the current behaviour. I'll give that a try on top of that PR to see if it works well and then we can merge once this is released!

emillon:
I'll have a look at that mdx failure.

NathanReb:
This is likely to just be a change in how opam prints the file, in which case I think we can just promote the change, the order of the fields doesn't really matter here!

emillon:
No, there's something going on with how the ellipsis is interpreted. Trying to reduce now!

NathanReb:
Ah yeah my bad! To be honest these cram tests are rather fragile and don't bring in much value. We should simplify the meaningful ones and get rid of the others.

Let me know if you find away around that ellipsis failure otherwise I'll rewrite that tests!

emillon:
After some debugging, it's a bug in mdx where EOL-ellipses can not be used after whole-line ellipses. I'll file that against mdx for completeness but the best solution is indeed to migrate to postprocessing. We don't have `jq` for opam files, so here's a solution using sed: (I can't push to your branch but this works):

```diff
-    $ cat _build/whatever.0.1.0/opam
+    $ cat _build/whatever.0.1.0/opam |\
+    > sed -e 's/\(sha...\)=[0-9a-f]*/\1/' \
+    >     -e 's/\(x-commit-hash:\) "[0-9a-f]*"/\1/'
```

emillon:
I think it's better to omit the ellipsis (`...`) in the output to make sure plain string matching is used.

kit-ty-kate:
Fair enough. Done!

emillon:
See https://github.com/realworldocaml/mdx/issues/329

kit-ty-kate:
All green and ready to go

kit-ty-kate:
cc @emillon @NathanReb

PR #348, opened 2021-02-23
Fix the process of inferring user name and repo

pitag-ha:
This commit fixes the following two minor bugs in the workflow of inferring the user name and repo from the opam file fields `homepage` or `dev-repo`.

- It adds a test case for the case the user puts a local path or an empty string into the opam file fields. Before the commit, if the user did so, the first two directories of the corresponding absolute path got inferred as user and repo.

- Before this commit, the uri's in `homemapge` and `dev-repo` fields were transformed into the corresponding uri's to the distribution by appending "releases/download/{version}/{project name}-{version}.tbz" to them. From these distribution uri's, user name and repo were inferred. 
Example: if the uri in the opam file fields was "https://foo.com/", then the process of inferring user name and repo would not return an error, but infer "releases" and "downloads" as user and repo, respectively. With this commit user and repo get inferred directly from the uri's without first transforming them into the distribution uri's.

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -16,11 +16,11 @@ let uri_domain uri =
 let uri_sld uri =
   match uri_domain uri with _ :: sld :: _ -> Some sld | _ -> None
 
-let uri_append u s =
-  match String.head ~rev:true u with
-  | None -> s
-  | Some '/' -> strf "%s%s" u s
-  | Some _ -> strf "%s/%s" u s
+let uri_append_to_base ~appendix base =
+  match String.head ~rev:true base with
+  | None -> appendix
+  | Some '/' -> strf "%s%s" base appendix
+  | Some _ -> strf "%s/%s" base appendix
 
 let chop_ext u =
   match String.cut ~rev:true ~sep:"." u with None -> u | Some (u, _) -> u
@@ -216,9 +216,6 @@ let dev_repo p =
   | None -> Ok None
   | Some r -> Ok (Some (chop_git_prefix r))
 
-let err_not_found () =
-  R.error_msg "no distribution URI found, see dune-release's API documentation."
-
 let dev_repo_is_on_github p =
   opam_field_hd p "dev-repo" >>| function
   | None -> false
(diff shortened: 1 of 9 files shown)


lib/pkg.ml
```diff
@@ -242,48 +239,59 @@ let path_of_distrib p =
...
-        | Some ("", path) -> "https://github.com/" ^ chop_ext path
-        | _ -> (
-            match String.cut ~sep:"git+ssh://git@github.com/" dev_repo with
-            | Some ("", path) -> "https://github.com/" ^ chop_ext path
-            | _ -> chop_git_prefix (chop_ext dev_repo))
-      in
-      path_of_distrib p >>| fun path -> Some (uri_append dev_repo path)
+let uri_of_dev_repo p =
+  opam_field_hd p "dev-repo"
+  >>| Stdext.Option.map ~f:(fun dev_repo ->
```
gpetiot, line 253: Would it be possible to move all this logic into another module uri.ml, that can just be fed the values of the fields, like `opam_field_hd p "dev-repo" >>| Stdext.Option.map ~f:Uri.of_dev_repo`, and same for other functions, to only keep the opam fields part in this file?

lib/pkg.ml
```diff
@@ -16,7 +16,7 @@ let uri_domain uri =
 let uri_sld uri =
   match uri_domain uri with _ :: sld :: _ -> Some sld | _ -> None
 
-let uri_append u s =
+let uri_append s u =
```
gpetiot, line 19: I think this would be the right time to name one argument as soon as the argument order becomes non trivial

gpetiot (approved):
This is a good catch, thanks for fixing this! The new module I suggested is non-blocking, I just think it would make the code more readable, but this is already a big improvement.

lib/pkg.ml
pitag-ha: Done. Thanks!

lib/pkg.ml
pitag-ha: That's a good idea! As discussed, I've done that in a [separate PR](https://github.com/ocamllabs/dune-release/pull/349) though since that's not directly related with the bug fix here.

PR #349, opened 2021-02-24
Clean up pkg.ml and extract a new module Uri

pitag-ha:
cc @gpetiot

### DIFF
--- a/lib/uri.mli
+++ b/lib/uri.mli
@@ -1,0 +1,27 @@
+val get_sld : string -> string option
+(** Get the URI's second level domain, if it has one. *)
+
+val append_to_base : rel_path:string -> string -> string
+(** Append a relative path to a base URI. *)
+
+val chop_git_prefix : string -> string
+(** Chop the prefix [git+] from a URI, if any. *)
+
+val to_https : string -> string
+(** Convert [git@] and [git+ssh://] into https URI's. Leave other URI's as is. *)
+
+module Github : sig
+  val to_github_standard : string -> (string, [> Bos_setup.R.msg ]) result
+  (** Convert a github pages URI into its correspondent repo URI. Leave other
+      URI's as is. *)
+
+  val get_user_and_repo :
+    string -> (string * string, [> Bos_setup.R.msg ]) result
+  (** Retrieve user name and repository name from a standard github URI. *)
+
+  val split_doc_uri :
+    string -> (string * string * Fpath.t, [> Bos_setup.R.msg ]) result
+  (** Parse a github pages URI of the form $SCHEME://$USER.github.io/$REPO/$PATH
+      into (user, repo, dir), where user=$USER and repo and dir are deduced from
+      $PATH *)
+end
--- a/tests/lib/test_uri.ml
+++ b/tests/lib/test_uri.ml
@@ -1,0 +1,94 @@
+let test_to_https =
+  let make_test ~input ~expected =
+    let name = "Uri.Github.to_https " ^ input in
+    let actual = Dune_release.Uri.to_https input in
+    let test_fun () = Alcotest.(check string) name expected actual in
+    (name, `Quick, test_fun)
+  in
+  [
+    make_test ~input:"git@github.com:user/repo.git"
+      ~expected:"https://github.com/user/repo";
+    make_test ~input:"git+ssh://git@github.com/user/repo"
+      ~expected:"https://github.com/user/repo";
+    make_test ~input:"git+https://gitlab.com/user/repo"
+      ~expected:"https://gitlab";
+    make_test ~input:"my_homepage.com" ~expected:"my_homepage";
+  ]
+
+let test_to_github_standard =
+  let make_test ~input ~expected =
+    let name = "Uri.Github.to_github_standard " ^ input in
+    let actual = Dune_release.Uri.Github.to_github_standard input in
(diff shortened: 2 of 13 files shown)


lib/uri.mli
```diff
@@ -0,0 +1,24 @@
+val get_sld : string -> string option
+(** Get the URI's second level domain, if it has one. *)
+
+val append_to_base : rel_path:string -> string -> string
+(** Append a relative path to a base URI. *)
+
+val chop_git_prefix : string -> string
+(** Chop the prefix "git+" from a URI, if there. *)
```
gpetiot, line 8: I would say "if any"

lib/uri.mli
```diff
@@ -0,0 +1,24 @@
...
+
+val to_https : string -> string
+(** Convert "git\@" and "git+ssh://" into https URI's. Leave other URI's as are. *)
```
gpetiot, line 11: I've looked this up and it seems like the "as is" expression is a fixed expression, so we should write either "as is" or "as they are" (slight preference for the shorter one)

gpetiot: Instead of writing `"git\@"` (I guess it's an ocamlformat bug that rewrites `"git@"` into `"git\@"`), we should write `[git@]`, the `[]` notation preserves the input better

lib/uri.mli
pitag-ha: I think my English might not be good enough to understand why to say "if any" in this context. Does "if present" also sound good to you? If not, I can change it to "if any".

lib/uri.mli
pitag-ha: Oh, that's interesting. I didn't know about the "as is" fixed expression. And also thanks for the ocamlformat tip!

lib/uri.mli
gpetiot: Yes I think "if present" also works, maybe I'm wrong but I think "if any" is really common to say "if this thing exists"

lib/uri.mli
```diff
@@ -0,0 +1,24 @@
...
+val append_to_base : rel_path:string -> string -> string
+(** Append a relative path to a base URI. *)
+
+val chop_git_prefix : string -> string
+(** Chop the prefix [git+] from a URI, if present. *)
+
+val to_https : string -> string
+(** Convert [git@] and [git+ssh://] into https URI's. Leave other URI's as is. *)
+
+val to_github_standard : string -> (string, [> Bos_setup.R.msg ]) result
```
NathanReb, line 13: I think this and the next two functions should probably live a `Github` submodule so we can clarify that all those functions only apply to github URIs.

In the future we can extract  type and use a better abstraction than strings or `Uri.t` for those.

NathanReb (commented):
Didn't we have unit tests for those functions? That might be a good occasion to extract some!

NathanReb:
We need to fix the tests on alpine. From a quick look the failure comes from the fact that it uses a more recent version of `git` and therefore warns about the default `master` branch name.

This should be fixed by adding a `-b main` to all our `git init` invocations, probably worth doing that in a separate PR though as it is unrelated to these changes!

gpetiot:
> We need to fix the tests on alpine. From a quick look the failure comes from the fact that it uses a more recent version of `git` and therefore warns about the default `master` branch name.
> 
> This should be fixed by adding a `-b main` to all our `git init` invocations, probably worth doing that in a separate PR though as it is unrelated to these changes!

The CI should be green upon rebasing on master I think!

lib/uri.mli
pitag-ha: Ok, I've used "if any". That way I get used to it.

lib/uri.mli
pitag-ha: > I think this and the next two functions should probably live a Github submodule so we can clarify that all those functions only apply to github URIs.

Yes, I agree, that's better.

> In the future we can extract type and use a better abstraction than strings or Uri.t for those.

That's what @gpetiot did in #215, right? I liked that PR.

pitag-ha:
> Didn't we have unit tests for those functions? That might be a good occasion to extract some!

No, we didn't have any. I've written some now. They just test the current behavior (they don't question it).

lib/uri.mli
NathanReb: Yep! That's definitely something we should aim at but I think this PR is already a huge improvement as it is.

Once we're heading towards 2.0 it'll be clearer how to refactor these and also hopefully much simpler since we won't have to parse github.io URIs for instance!

NathanReb (approved):
Awesome tests, glad we added those!

Thanks!

tests/lib/test_uri.ml
```diff
@@ -0,0 +1,94 @@
...
+      ~expected:(Ok ("user", "repo"));
+  ]
+
+let test_split_doc_uri =
+  let make_test ~input ~expected =
+    let name = "Uri.Github.split_doc_uri " ^ input in
+    let actual = Dune_release.Uri.Github.split_doc_uri input in
+    let test_fun () =
+      Alcotest.(
+        check (Alcotest_ext.result_msg (triple string string Alcotest_ext.path)))
```
emillon, line 73: Hi, just FYI this requires alcotest >= 1.3.0. Can you bump the dep?

emillon: (this = `triple`)

PR #351, opened 2021-03-17
Add dune project check

pitag-ha:
With this commit, `check` also checks if the dune project contains a name stanza. That's necessary for running `dune subst` when constructing the tarball. It might also become necessary in 2.0.0 to infer the main package name. At the moment, for inferring the main package name it's sufficient.

### DIFF
--- a/lib/pkg.mli
+++ b/lib/pkg.mli
@@ -169,6 +169,12 @@ val version_line_re : Re.t
 val prepare_opam_for_distrib :
   version:string -> content:string list -> string list
 
+(** {1 Dune project} *)
+
+val dune_project_name : Fpath.t -> (string option, [> Rresult.R.msg ]) result
+(** Returns the name stanza entry of the dune project, if any; else, returns
+    `None`. Returns an error, if one of the system calls fails. *)
+
 (**/**)
 
 (*---------------------------------------------------------------------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@
   `--draft` for 'dune-release publish' and 'dune-release opam submit' commands.
   (#248, @gpetiot)
 - Add a new command `check` to check the prerequisites of dune-release and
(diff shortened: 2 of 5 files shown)


lib/pkg.mli
```diff
@@ -169,6 +169,12 @@ val version_line_re : Re.t
 val prepare_opam_for_distrib :
   version:string -> content:string list -> string list
 
+(** {1 Dune project} *)
+
+val dune_project_name : Fpath.t -> (string option, [> Rresult.R.msg ]) result
+(** Returns the name stanza entry of the dune project, if any; else, returns
+    `None`. Returns an error, if one of the system calls fails. *)
```
pitag-ha, line 176: I don't really like exposing even more functions in `Pkg`, particularly one that doesn't even depend on some `Pkg.t`. What do you two think? Is that ok or would you move that somewhere else?

gpetiot: I thought there already was a library to read dune-project files, well I looked for it but couldn't find anything, maybe it's planned to get it out of dune but wasn't done yet

pitag-ha: Do you mean dune itself exposing its library to parse the dune project or an independent library?

gpetiot: I was thinking of an independent library

gpetiot: Maybe it can make sense to put this function in a dune_project.ml file, so we can put all similar functions there if ever we do more checks on dune-project files in the future

gpetiot (approved):
Looks good to me!

lib/pkg.mli
NathanReb: This fonction will be useful in the future I think. We should split `Pkg` into `Pkg` and `Repo`. It could fit the latter.

In the meantime I think it's okay to have it here!

NathanReb (approved):
This looks good! Thanks!

PR #352, opened 2021-03-18
Deprecate opam file format 1.x support as well

NathanReb:
Previously only the support for the opam client was deprecated.


gpetiot (approved):
Looks good to me, thanks! dune-release 2.0.0 will be a breeze of fresh air!

PR #358, opened 2021-04-20
Fix inference of gh URIs

NathanReb:
This PR fixes a bug we recently introduced where the logic to extract a Github repository when needed was broken and would only consider the homepage in most cases instead of falling back to the `dev-repo` when `homepage` wasn't on github.

The fact that we were simply passing strings around wasn't helping so I extracted a type for github URIs (that we could in the future turn as representation for github repos really).

This fixes the bug I was running into when testing the latest dune-release before attempting to release 1.5.

While doing this work I realized there were a few things we were doing wrong.

The first one was that we were trying to parse the user provided `--distrib-uri` as a github URI which is inherently wrong since the fact that the user needs to provide the distrib-uri on its own most likely means he's not releasing on github. I fixed that by removing this logic entirely as in those cases it was clear we were dependent on the repo being hosted on github so we're now using the dev-repo/homepage directly. I fixed this here as it was logical and simpler to do so as part of the refactoring.

Another bug that I will fix in a subsequent PR is that `opam submit` will fail if we can't find a github repo associated with the package. It shouldn't be the case and non github users should still be able to release using dune-release at this point.

Let me know what you think!


NathanReb:
Hmm as a matter of fact, I might rename `Github_uri` to `Github_repo` as that's what we wish to represent here.

lib/delegate.ml
```diff
@@ -34,7 +34,10 @@ let publish_distrib ?token ?distrib_uri ~dry_run ~msg ~archive ~yes ~draft pkg =
       Pkg.tag pkg >>= fun version ->
       (match distrib_uri with
       | Some uri -> Ok uri
-      | None -> Pkg.infer_distrib_uri pkg)
+      | None ->
+          (* This is an absolute non-sense but we're keeping it to preserve the
+             behaviour until 2.0 and removal of delegates *)
+          Pkg.infer_github_distrib_uri pkg)
```
gpetiot, line 40: That would be a slight behavior change but we could require a `--dist-uri` when using a delegate, the user would get a more precise error message.

lib/stdext.mli
```diff
@@ -40,12 +40,32 @@ module Unix : sig
...
+  val pp :
+    (Format.formatter -> 'a -> unit) -> Format.formatter -> 'a option -> unit
+
+  val equal : ('a -> 'a -> bool) -> 'a option -> 'a option -> bool
+
   val map : f:('a -> 'b) -> 'a option -> 'b option
 
   val bind : f:('a -> 'b option) -> 'a option -> 'b option
+
+  module O : sig
```
gpetiot, line 52: I would call the module `Monad_infix` or `Infix`

dune-release.opam
```diff
@@ -30,6 +30,7 @@ depends: [
   "alcotest" {with-test}
   "mdx" {with-test & >= "1.6.0"}
   "yojson"
+  "uri"
```
gpetiot, line 33: This is not used

lib/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (name dune_release)
  (public_name dune-release)
- (libraries fmt bos curly opam-state rresult bos.setup yojson))
+ (libraries fmt bos curly opam-state rresult bos.setup uri yojson))
```
gpetiot, line 4: `uri` is not used

lib/stdext.mli
NathanReb: The `O` stands for `Operators`, it's a convention I got from dune IIRC, I like it as it can be used for any kind of operators, not only infix ones.

dune-release.opam
NathanReb: Good catch!

lib/delegate.ml
NathanReb: Yeah I think it would make sense but it should probably be done in a separate PR, what do you think?

lib/stdext.mli
gpetiot: Alright!

lib/delegate.ml
gpetiot: Yes it's probably better indeed

gpetiot (approved):
Looks good to me, thanks!

NathanReb:
@gpetiot could you try this on a dummy package or even on ocamlformat see if it improves the situation and/or if you can spot any bug that isn't there on master? I tested myself on a package of mine but I'd prefer to err on the safe side here!

gpetiot:
> @gpetiot could you try this on a dummy package or even on ocamlformat see if it improves the situation and/or if you can spot any bug that isn't there on master? I tested myself on a package of mine but I'd prefer to err on the safe side here!

I keep hitting this issue:
```sh
$ dune-release opam submit
[-] Submitting 
[-] Preparing pull request to ocaml/opam-repository
[-] Fetching https://github.com/ocaml/opam-repository.git#master
[-] Checking out a local release-test-dune-release-v0.0.0 branch
dune-release: [ERROR] /home/gpe/opam-repository/_build/test-dune-release.0.0.0 does not exist, did you run:
    dune-release opam pkg -p test-dune-release
```
although `dune-release opam pkg` has been run before, and same with the full-auto mode

(my conf: `remote: https://github.com/gpetiot/opam-repository.git`
`local: /home/gpe/opam-repository`)

NathanReb:
Yeah that's the next bug on my todo!

@pitag-ha have you had a chance to take a look/test this?

bin/opam.ml
```diff
@@ -197,27 +197,25 @@ let submit ?distrib_uri ~token ~dry_run ~yes ~opam_repo ~user local_repo
   list_map Pkg.name pkgs >>= fun names ->
   let title = strf "[new release] %a (%s)" (pp_list Fmt.string) names version in
   Pkg.publish_msg pkg >>= fun changes ->
-  (match distrib_uri with Some uri -> Ok uri | None -> Pkg.infer_repo_uri pkg)
-  >>= Github_uri.get_user_and_repo
-  >>= fun (distrib_user, repo) ->
+  Pkg.infer_github_repo_uri pkg >>= fun { owner; repo } ->
```
pitag-ha: If I'm not missing something, this is a tiny breaking change: someone runs
```
dune-release opam submit --dist-uri <some_github_uri>
```
Before, that used to work. Now, it only works, if the distribution uri can be inferred from the opam file. Is that right? Do you think that can be ignored?

pitag-ha (commented):
Thanks, that makes that part so much more straight-forward!

> This PR fixes a bug we recently introduced where the logic to extract a Github repository when needed was broken and would only consider the homepage in most cases instead of falling back to the `dev-repo` when `homepage` wasn't on github.

I'm not being able to reproduce that. For me it works both on master and on this branch, if `homepage` doesn't exist, but `dev-repo` does. Is that expected?

NathanReb:
> I'm not being able to reproduce that. For me it works both on master and on this branch, if `homepage` doesn't exist, but `dev-repo` does. Is that expected?

Yeah, to trigger the bug you have to have a `home` field that points to something else than github! Try with a `dummy.com` URL for instance, you should see the bug then!

bin/opam.ml
NathanReb: I think this is very very unlikely and that this is inherently broken. To clarify, for that to happen we would need a user whose repo is not hosted on github that happened to have uploaded the distrib tarball somewhere on github. He passes that URI which we then use to determine a user and repo (which won't correspond to the one where the project is hosted, since it's not on github). We then either use them to rewrite the publication message in a github specific way and might also use that to open a PR to opam-repo if they have a broken configuration.

My impression is that would never work because it's using the wrong information in the wrong place.

The only case where it might eventually work is if the project is mirrored on github but even then, I'm not sure it would do the right thing. If we want to support mirroring on github (which we probably should) we'd need an CLI option so one can have the original repo in their opam file but pass the github mirror URI to dune-release so it release to the mirror as it would for a regular github hosted project.

To be perfectly honest I'm not 100% sure myself that I considered all the options but I feel like this has never worked for anyone. Do you have an example in mind of a situation where one would want this behaviour?

PR #359, opened 2021-04-26
Fix opam submit for non gh projects

NathanReb:
Depends on #358

`dune-release opam submit` used to fail when releasing a package that wasn't hosted on github even though this is not required to go through this step.
One should be able to upload the distrib tarball to a custom place on the web and then still use dune-release to create the opam-repository PR as long as they provide the right informations on the command line.

This PR fixes this by skipping the github specific step that consists of rewriting the PR/issue references to the cross repo reference format in the PR message if neither of the `home` or `dev-repo` fields in the opam file points to github.

While working on this I also discovered there were some inconsistencies in the way the configuration is handled so I will fix them in a subsequent PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -49,6 +49,8 @@
 
 ### Fixed
 
+- Fix a bug in `opam submit` preventing non-github users to create the opam-repo PR
+  via dune-release. (#359, @NathanReb)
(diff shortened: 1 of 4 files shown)


gpetiot (approved):
It looks good to me! (didn't test it yet as the change looks pretty straight forward)

PR #360, opened 2021-04-26
Improve documentation of the --dist-uri option

NathanReb:
I wanted to emphasize that this is useful for user that do not release to github and that other users should probably not have to care about it!


gpetiot (approved):
Looks good to me!

PR #361, opened 2021-04-26
Fix opam submit

NathanReb:
We recently introduced a bug that completely broke the command as it wasn't able to properly locate the folder to add to opam-repo for the release.

This was introduced very recently so there's no need for a changelog entry!


NathanReb:
@gpetiot this should allow you to try dune-release on ocamlformat again so we can focus on the multi-opam related problems if they persist!

gpetiot (approved):

Issue #362, opened 2021-04-28
custom opam repositories are not well-handled


mimoo:
Hello hello!

Dune-release is really a fantastic piece of software, I ran into some trouble today that I managed to solve but this might help someone else.

When using a custom opam repository, you can't use `dune-release` directly, you have to follow [each step](https://ocaml-explore.netlify.app/images/dune-release.png) in order until the last `dune-release opam submit`, which you have to write as `dune-release opam submit

NathanReb:
Thanks for reporting this!

We should indeed add the `--opam-repo` to the default command, that's an omission on our part.

That being said, the rest is not exactly true, I for instance do release to a custom opam repository quite often when testing dune-release and do not experience any of the problem you mention. You should not delete your `release.yml` config file, just properly set it up the first time. The `remote` field should point to your own fork (or one you have write access to) of `ocaml/opam-repository` or the opam repository you want to release to. The `local` field should point to your local clone of that repo or fork, again dune-release should pull and push to the right places (i.e. pull from `--opam-repo` and push to your config's `remote`).

If your configuration is set up this way, it should all work fine. One thing that might happen is that the opam repository you want to release to is not based on `ocaml/opam-repository`, in which case you might not be able to share the same fork for releases to the custom and the regular opam-repo.

If that's the case I'd advise you set the `--local-repo` and `--remote-repo` CLI options accordingly to override your configuration rather than delete the config file everytime.

The key point here probably is the lack of documentation for dune-release and we're well aware of that. It might also be the case that these kind of workflows are not perfectly well supported or rather that you'll encounter bugs when trying more custom release workflows. We've been working to improve the experience for all users of dune-release lately so I expect things are getting better and better but there's still work to do on that front so don't hesitate to report any rough edges or bugs.

We've also considered adding some kind of per-repo configuration so that when you want to release different repos differently you don't have to pass a whole bunch of CLI options for that to work. This is not there yet and we unfortunately have higher priority things to fix and improve so in the meantime I'd suggest you have a small script that invokes dune-release with the right set of option in your repo for that purpose, keeping your most common `remote` and `local` in your global config file.

NathanReb:
It's also worth noting that we recently discovered bugs and inconsistencies in the configuration and the way it is handled by the tool so expect improvements on that front as well!

mimoo:
> just properly set it up the first time

[...]

Issue #357, opened 2021-04-16
Improve consistency among flags


pitag-ha:
There are a couple of options that can be passed to subcommands such as `distrib`, but not to the all-in-one `dune-release`/`dune-release bistro` command. For example the `--skip-<build/tests/lint>` flags.

I suggest adding those options to the `dune-release` command for consistency and to increase the usability of that command.

PR #363, opened 2021-04-29
Add missing options to the bistro command

NathanReb:
Some of the `opam submit` specific options weren't allowed to be passed to the default command, preventing it from being used for releases to custom repositories among other use cases.

This should help with #362 !

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


gpetiot (approved):
LGTM!

pitag-ha:
Looks good! There are also a couple of other flags from other commands such as `distrib` that are missing in `bistro`. See #357.

PR #364, opened 2021-04-29
Fix Bistro command

NathanReb:
Because of how we rewrote it, it seems that it was evaluating the commands in the wrong order.


gpetiot (approved):
Good catch!

PR #366, opened 2021-05-04
Keep the first name encountered in dune-project.

mefyl:
When trying to infer the name from the `dune-project` file, we currently keep the last `(name ...)` encountered. However the root "name" stanza is usually around the beginning on the file, and it should probably take precedence on "package" stanza names.

I suppose we don't want to include a more advanced sexp parser to keep dependencies to a bare minimum.

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -291,8 +291,8 @@ let dune_project_name dir =
       List.fold_left
         (fun acc line ->
           (* sorry *)
-          match String.cut ~sep:"(name " (String.trim line) with
-          | Some (_, s) ->
+          match (acc, String.cut ~sep:"(name " (String.trim line)) with
+          | None, Some (_, s) ->
               Some
                 (String.trim
                    ~drop:(function ')' | ' ' -> true | _ -> false)


Leonidas-from-XIV:
@NathanReb Similar to @gpetiot's recent YAML PR (#355) I would suggest instead of trying to parse the sexp by hand to depend on `sexplib` and parse this in a way that is less likely to get confused by e.g. reformatting. Sexplib has -despite being part of the Core suite- a rather slim dependency cone and does not introduce C dependencies, so it should not be too painful of an addition. What do you think?

gpetiot:
In the meantime, maybe we can just consider that the line starting with `(name` (_without indentation_) is the one we're after, since the package names should be indented with at least 1 space. It sounds more robust than just picking the first one.

This bug is really annoying when releasing ocamlformat.

Leonidas-from-XIV:
@gpetiot We discussed this with @NathanReb and another possibility would be to just use the name of the repo. The current behaviour is a bit strange anyway, leading to e.g. `ocaml-lsp` being called `jsonrpc`.

Leonidas-from-XIV:
Closed by #445

PR #368, opened 2021-05-19
Improve GH token management

NathanReb:
The CLI option and env var used to take path which made them unusable in a CI context. They now take the token value directly.

This also improves how the token is passed around, moving the config reading bits as close to the CLI entry points as possible so the rest of the code doesn't have to deal with this and simply require a mandatory token argument, with the value and not the path.

This finally fixes a bug where the content of token files wasn't always sanitized.

### DIFF
--- a/lib/config.mli
+++ b/lib/config.mli
@@ -29,7 +29,12 @@ val v :
   Pkg.t list ->
   (t, Bos_setup.R.msg) result
 
-val token : dry_run:bool -> unit -> (Fpath.t, Bos_setup.R.msg) result
+val token :
+  ?cli_token:string -> dry_run:bool -> unit -> (string, Bos_setup.R.msg) result
+(** Returns the token value that should be used for github API requests. If a
+    [cli_token] was provided, it is returned. Otherwise the token file in the
+    config dir is looked up. If it exists, its content is returned, if it does
+    not, the user is prompted for a token which will be then saved to that file. *)
 
 val keep_v : bool -> (bool, Bos_setup.R.msg) result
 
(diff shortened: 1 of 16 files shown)


NathanReb:
I tested the following scenarios on https://github.com/NathanReb/dune-release-testing:
1. A release using individual subcommands with my existing config token
2. A release using the default command with my existing config token
3. A release using the default command with the config token file deleted and my token passed through the env var `DUNE_RELEASE_GITHUB_TOKEN` 
4. A release using the default command without a config token set. I followed the steps suggested in the logs to generate the token and pasted it into dune-release (note for later, this bit needs some feedback because it really feels like nothing is happening even though you did paste the token).

It all worked, as a result of `4` the token file is properly saved.

If you have any other meaningful test cases in mind please let me know, I'm happy to try them out to make sure everything works as intended.

lib/config.mli
```diff
@@ -29,7 +29,12 @@ val v :
   Pkg.t list ->
   (t, Bos_setup.R.msg) result
 
-val token : dry_run:bool -> unit -> (Fpath.t, Bos_setup.R.msg) result
+val token :
+  ?cli_token:string -> dry_run:bool -> unit -> (string, Bos_setup.R.msg) result
+(** Returns the token value that should be used for github API requests. If a
+    [cli_token] was provided, it is returned. Otherwise the token file in the
+    config dir is looked up. If it exists, it's content is returned, if it does
```
gpetiot, line 36: should be "its content"

gpetiot (approved):
I have no additional scenario to test in mind but the code looks good to me!

PR #369, opened 2021-05-21
Use proper token based authentication for github v3 API

NathanReb:
This changes from using `-u user:token` to authenticate Github v3 API request to using the recommended `-H "Authorization: token <token>"`.

This allows for better interaction with some organization hosted repos and also allow us to drop the user config field which we won't need anymore!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,8 @@
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


gpetiot (approved):
It looks good to me, thanks!
It should be easier to factorize the authentication for v3 and v4 APIs now.

PR #371, opened 2021-06-21
Fix long option names for local and remote repo

NathanReb:


NathanReb:
No changelog here since this was introduced recently!

gpetiot (approved):
Good catch

PR #372, opened 2021-06-21
Deprecate user options and config field and rely on remote-repo entirely

NathanReb:
This PR deprecates the `user` CLI option and config field.

The value, whether it comes from the CLI or the config is ignored, we rely entirely on remote-repo now.
Deprecation warnings are emitted whenever someone tries to use the CLI option and if they try to read or modify the field via the `config` subcommand.

New configuration will be created without a user field from now on.

I fixed the config creation step as before it would crash if the repo wasn't hosted on github because it was trying to parse the opam files field, looking for a github username to use as a suggestion in the config creation quiz.

There are further things to fix here, in particular, the fact that we still try to create a config even if all the information is provided via the command line. I will take care of this in a subsequent PR.

I have to test this a bit so I'm marking as draft but feedback is welcome!

### DIFF
--- a/bin/opam.ml
+++ b/bin/opam.ml
@@ -117,7 +117,7 @@ let pp_opam_repo fmt opam_repo =
   let user, repo = opam_repo in
   Format.fprintf fmt "%s/%s" user repo
 
-let open_pr ~dry_run ~changes ~remote_repo ~user ~branch ~token ~title
+let open_pr ~dry_run ~changes ~remote_repo ~fork_owner ~branch ~token ~title
     ~opam_repo ~auto_open ~yes ~draft pkg =
   Pkg.opam_descr pkg >>= fun (syn, _) ->
   Pkg.opam_homepage pkg >>= fun homepage ->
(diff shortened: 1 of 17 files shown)


NathanReb:
I just tested this and had no issue running it without an existing config, reusing the freshly created config or with my exisint config.

If you have any good test scenario in mind please let me know!

lib/config.ml
```diff
@@ -69,99 +69,107 @@ let of_yaml_exn str =
...
   in
   Fpath.of_string local >>= fun local ->
-  let v = strf "user: %s\nremote: %s\nlocal: %a\n" user remote Fpath.pp local in
-  OS.Dir.create Fpath.(parent file) >>= fun _ ->
-  OS.File.write file v >>= fun () ->
-  Ok
+  let config =
     {
-      user = Some user;
+      user = None;
```
gpetiot, line 154: Can we go further and remove this field in this PR to avoid any mis-use?

NathanReb: We probably could but I'd rather err on the safe side here, simply going through deprecation first, removing it in 2.0.

bin/opam.ml
```diff
@@ -208,30 +220,16 @@ let submit ~token ~dry_run ~yes ~opam_repo ~user local_repo remote_repo pkgs
     | Some { owner; repo } -> Text.rewrite_github_refs ~user:owner ~repo changes
     | None -> changes
   in
-  let user =
-    match user with
-    | Some user -> Ok user (* from the .yaml configuration file *)
-    | None -> (
-        match Github.Parse.user_from_remote remote_repo with
```
gpetiot, line 215: All uses of this functions have been removed, we can remove it from github.ml

gpetiot:
> I fixed the config creation step as before it would crash if the repo wasn't hosted on github because it was trying to parse the opam files field, looking for a github username to use as a suggestion in the config creation quiz

Can we have a cram test for this scenario? (the ones where we are just using the deprecated options are not really worth it in my opinion)

NathanReb:
> > I fixed the config creation step as before it would crash if the repo wasn't hosted on github because it was trying to parse the opam files field, looking for a github username to use as a suggestion in the config creation quiz
> 
> Can we have a cram test for this scenario? (the ones where we are just using the deprecated options are not really worth it in my opinion)

To be honest I'm not sure how to easily test this since it relies on the absence of the configuration file and will go into interactive mode. If you have a good suggestion I'd gladly take it but I'm afraid we'd end up with yet another fragile cram test.

bin/opam.ml
NathanReb: Good catch!

gpetiot (approved):

PR #373, opened 2021-06-29
Fix interpretation of --local-repo and --remote-repo

NathanReb:
These options were properly used but they didn't prevent dune-release from accessing the configuration file and eventually, from
creating a new one through interactive prompts. This behaviour prevented it from being used in a CI.

Now the configuration is only looked up (and eventually created) if dune-release doesn't have enough information to proceed with
the release.

In order to fix this I did a few things:
- I made both config fields mandatory, dune-release can't produce a config without those filled up anymore. This used to be pretty much the case but one could manage to do it if they created the config using `dune-release config set`.
- I replaced `Config.v` with an accessor like function, similar to what we use for `keep_v` and `auto_open`. If the CLI provided value is provided, it's used, if we lack information, we look in the config for it. The difference to the above mentioned flags is that we don't have a default value for the opam fork so if there is no config, we start the creation quizz.

The creation quizz also doesn't use the values passed on the CLI as suggestions, this didn't seem sensible to me as the main usecase for using those options is either to run dune-release in a script or a CI or to override the values in the config file.

I also added a `dune-release config create` so one can access the config creation quizz without having to go through an almost full release.

One thing worth noting is that it breaks the behaviour of `dune-release config` when there's no configuration. Before it would either set a single field or display all fields as `<unset>`. I considered this to be okay as I'm not sure anyone was relying on this behaviour but I'm happy to reconsider if you think this is not the right behaviour or that it's too likely to break someone's workflow!

### DIFF
--- a/lib/config.ml
+++ b/lib/config.ml
@@ -22,14 +22,12 @@ end
 
 type t = {
   user : string option;
-  remote : string option;
-  local : Fpath.t option;
+  remote : string;
+  local : Fpath.t;
   keep_v : bool option;
   auto_open : bool option;
 }
 
-let empty =
-  { user = None; remote = None; local = None; keep_v = None; auto_open = None }
 
 let of_yaml_exn str =
   (* ouch *)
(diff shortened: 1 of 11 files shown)


NathanReb:
As usual, I'm marking as draft so I can run a bit of manual testing but your feedback is already appreciated!

Also if you come up with a specific thing to test, it's always welcome!

NathanReb:
This is ready for review!

lib/config.ml
```diff
@@ -255,6 +247,20 @@ let keep_v v = if v then Ok true else read (fun t -> t.keep_v) false
 
 let auto_open v = if not v then Ok false else read (fun t -> t.auto_open) true
 
+let opam_repo_fork ?pkgs ~remote ~local () =
```
gpetiot, line 250: I think it would be nice to find a way to distinguish the values that are only the CLI options and the ones that are the unified values (CLI options / file config), because it's a bit confusing to have `local`/`remote` values in input and in output.

gpetiot (approved):
Minor comment that would make the code easier to follow, but it looks good to me anyway, thanks!

lib/config.ml
NathanReb: Yeah I agree, do you have something in particular in mind? Would renaming the labels be enough for instance or do you have something more involved in mind?

gpetiot: I was thinking of something more involved (to be done in a separate PR), like having a type `'a t` in the cli for all options, same for values returned in config.ml, and then unification (not sure it's the right term) functions would have type like: `string Cli.t -> string Config.t -> string`

NathanReb: Yeah that's a good idea! I can give it a quick try. What would you think about this:

Since it only matters for config, I can extract just a `Config.Cli.t` type with a smart constructor, keeping it abstract. I then use this constructor for the CLI args `keep_v`, `no_auto_open`, `local_repo` and `remote_repo`. The only way to consume those then is to pass them to the right functions in `Config`.

That's fairly small so I can include it in this PR and it feels like it would solve our problem here.

NathanReb: Ok I'm working on this but the change is slightly more involved than I hoped it would be so let's deal with it in a separate PR.

PR #374, opened 2021-07-01
Add --no-auto-open to default command and improve config CLI args mgmt

NathanReb:
This PR adds the missing `--no-auto-open` flag to the default command.

It also slightly improves the handling of config CLI arguments, i.e. CLI args that are meant to override or replace values that might be present in the user's config file. The correspongind cmdliner terms now wrap `'a Config.Cli.t` values, which have to be passed to the corresponding `Config` functions in order to extract a usable value out of them. This module now centralizes the logic responsible for selecting the right value based on what was passed on the command line and what is present in the config.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- Add `--no-auto-open` to the default command. It was previously only available for
+  `dune-release opam`. (#374, @NathanReb)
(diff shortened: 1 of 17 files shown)


NathanReb:
I think I'll cut the release after this one!

gpetiot (approved):
Looks good to me, thanks! Having different types to distinguish where the value of the configuration bits come from is something we should have done long ago!

PR #377, opened 2021-07-07
Use yaml library to parse release.yml configuration file

gpetiot:
Rewriting the parsing of the release.yml file that was a bit adventurous.

This PR is a reboot of #355 that was deleted (and couldn't bring it back) since the master branch has been renamed.

### DIFF
--- a/lib/config.ml
+++ b/lib/config.ml
@@ -30,49 +30,46 @@ type t = {
 
 module Opam_repo_fork = struct
   type t = { remote : string; local : Fpath.t }
+
+  let err_msg =
+    Rresult.R.msg
+      "Dune-release configuration file is invalid: both local and remote field \
+       should be set"
+
+  let of_yaml_values dict =
+    let rec aux ((r, l) as acc) = function
+      | [] -> Ok acc
+      | ("remote", `String s) :: t -> aux (`Remote (Some s), l) t
+      | ("local", `String s) :: t ->
+          Fpath.of_string s >>= fun s -> aux (r, `Local (Some s)) t
+      | _ :: t -> aux acc t
+    in
+    aux (`Remote None, `Local None) dict >>= fun (`Remote r, `Local l) ->
+    Stdext.Option.to_result ~none:err_msg r >>= fun remote ->
+    Stdext.Option.to_result ~none:err_msg l >>= fun local ->
+    Ok { remote; local }
 end
 
-let of_yaml_exn str =
-  (* ouch *)
-  let lines = String.cuts ~empty:false ~sep:"\n" str in
-  let dict () =
-    List.map
-      (fun line ->
-        match String.cut ~sep:":" line with
-        | Some (k, v) -> (String.trim k, String.trim v)
-        | _ -> failwith "invalid format")
-      lines
-  in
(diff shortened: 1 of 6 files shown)


lib/config.ml
```diff
@@ -30,49 +30,42 @@ type t = {
...
+        | ("remote", `String _) :: t -> aux acc t
+        | ("local", `String _) :: t -> aux acc t
+        | ("auto-open", `Bool b) :: t -> aux { acc with auto_open = Some b } t
+        | ("keep-v", `Bool b) :: t -> aux { acc with keep_v = Some b } t
+        | (key, _) :: _ ->
+            R.error_msgf "%S is not a valid configuration key." key
+      in
+      Opam_repo_fork.of_yaml_values dict >>= fun { remote; local } ->
+      aux { user = None; remote; local; auto_open = None; keep_v = None } dict
+  | _ -> R.error_msg "invalid format"
```
Leonidas-from-XIV, line 68: It would be good to have a better error message here, something like `Object expected, got XYZ`.

lib/config.ml
```diff
@@ -30,49 +30,42 @@ type t = {
...
-  {
-    user = find "user";
-    remote;
-    local;
-    auto_open = find_b "auto-open";
-    keep_v = find_b "keep-v";
-  }
+  let err_msg =
+    Rresult.R.msg
+      "Dune-release configuration file is compromised: both local and remote \
```
Leonidas-from-XIV, line 36: I know the wording was like this before but "compromised" sounds really misleading here, maybe "invalid" would be better.

lib/config.ml
```diff
@@ -30,49 +30,42 @@ type t = {
...
+       field should be set"
+
+  let of_yaml_values dict =
+    let rec aux ((`Remote r, `Local l) as acc) = function
```
Leonidas-from-XIV, line 40: If I see this right there is not need to unpack the polymorphic variants, since they're only being reconstructed later on.

lib/config.ml
```diff
@@ -158,11 +151,6 @@ let create ?pkgs () =
   get_path () >>= fun file ->
   create_config ?pkgs file >>= fun _ -> Ok ()
 
-let find () =
-  get_path () >>= fun file ->
-  OS.File.exists file >>= fun exists ->
-  if exists then OS.File.read file >>= of_yaml >>| fun x -> Some x else Ok None
-
```
Leonidas-from-XIV, line 165: Nice deduplication :+1:

Leonidas-from-XIV (commented):
Looks good. I have some minor comments. I kinda wish we used some simpler format than YAML that will get `ctypes` and C code into our dependency graph but this is a separate discussion and actually using an actual YAML parser is definitely better than hand-rolling a bad version of one.

lib/config.ml
gpetiot: done

lib/config.ml
gpetiot: done

lib/config.ml
gpetiot: done

Leonidas-from-XIV (approved):
Nice!

PR #383, opened 2021-09-06
Infer version from tag

### LINKED ISSUE #381: Wrong distrib name with dune-release 1.5.0
When releasing Merlin `dune-released` refused to sue the tag name as a version for the archive file and the release name. Instead it used the name `merlin` leading to the confusing archive name `merlin-merlin.tbz`.

```
$ dune-release tag v4.3-411
$ dune-release distrib
[-] Building source archive
[+] Wrote archive _build/merlin-merlin.tbz
```

With `dune-release` 1.4.0 I get the expected behavior:
```
$ dune-release tag v4.3-411
$ dune-release distrib
[-] Building source archive
[+] Wrote archive _build/merlin-v4.3-411.tbz
```

I wonder if that could come from undesirable tag detection since our changelog starts with:
```
merlin 4.3
----------
changes....
```

To reproduce, you can do a dry run on current merlin master.

Leonidas-from-XIV:
This changes the source from where to read the version from the changelog to the tag. To be able to serialize typical versions into git tags it uses a simplified convention of [DEP-14](https://dep-team.pages.debian.net/deps/dep14/): `~` becomes `_` and vice versa.

Checking opam-repository for "exotic" versions with

```sh
fd --type d '.+\..+' | rg '\.(.+)' --replace '$1' --only-matching | rg '[^-\.\d\w+~]''
```

Yields that once dots, digits, letters, `-`, `+` and `~` are excluded, there are no other surprising characters that we need to be careful in practice.

This also allows users of `dune-release` that aren't using the changelog file as source of truth use the tool and even maps tags like `1.3_beta3` into OPAM compatible `1.3~beta3` which could be quite useful.

Fixes #381

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,9 @@
 ### Fixed
 
 - Change the `---V` command option to be `-V` (#388, @Leonidas-from-XIV)
+- Infer release versions are inferred from VCS tags. This change allows using
+  `dune-release` on projects that do not use the changelog or have it in a
+  different format.  (#381, #383 @Leonidas-from-XIV)
 
 ### Removed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### Added
 
+- Added support for creating releases from unannotated Git tags. `dune-release`
+  supported unannotated tags in a few places already, now it supports using
+  them for creating a release. (#383, @Leonidas-from-XIV)
+
 ### Changed
 
 ### Deprecated
--- a/lib/vcs.ml
+++ b/lib/vcs.ml
@@ -209,34 +209,11 @@ let git_ls_remote ~dry_run r ~kind ~filter upstream =
 let git_submodule_update ~dry_run r =
   run_git_quiet ~dry_run r Cmd.(v "submodule" % "update" % "--init")
 
-let unallowed_substrings = Re.(compile (alt [ str "@{"; str ".." ]))
+(* See the reference here: https://git-scm.com/docs/git-check-ref-format
+   * Similar escape as DEP-14: https://dep-team.pages.debian.net/deps/dep14/ *)
+let git_escape_tag t = String.map (function '~' -> '_' | c -> c) t
 
-(* See the reference here: https://git-scm.com/docs/git-check-ref-format *)
-let git_escape_tag t =
-  let last = String.length t - 1 in
-  if String.equal t "@" then "_AT_"
-  else
-    String.fold_left
-      (fun (ret, i) c ->
-        let s =
-          match (i, c) with
-          | 0, '/' -> "_SLASH_"
-          | i, '/' when i = last -> "_SLASH_"
-          | i, '.' when i = last -> "_DOT_"
-          | _, ' ' -> "_SPACE_"
-          | _, '~' -> "_TILDE_"
-          | _, '^' -> "_CARET_"
-          | _, ':' -> "_COLON_"
-          | _, '?' -> "_QUEST_"
-          | _, '*' -> "_TIMES_"
-          | _, '[' -> "_LBRACKET_"
-          | _, '\\' -> "_BSLASH_"
-          | _ -> String.of_char c
-        in
-        (ret ^ s, i + 1))
-      ("", 0) t
-    |> fst
-    |> Re.replace_string ~all:true unallowed_substrings ~by:"__"
+let git_unescape_tag t = String.map (function '_' -> '~' | c -> c) t
 
 (* Hg support *)
 
@@ -436,6 +413,10 @@ let submodule_update ~dry_run r =
 let escape_tag = function
   | `Git, _, _ -> git_escape_tag
   | `Hg, _, _ -> hg_escape_tag
+
+let unescape_tag = function
+  | `Git, _, _ -> git_unescape_tag
+  | `Hg, _, _ -> fun x -> x
 
 (*---------------------------------------------------------------------------
    Copyright (c) 2016 Daniel C. Bünzli
--- a/lib/vcs.ml
+++ b/lib/vcs.ml
@@ -217,13 +217,12 @@ let git_unescape_tag t = String.map (function '_' -> '~' | c -> c) t
 
 (* Hg support *)
 
-(* Mercurial allows everything but :, \r and \n:
- * See https://www.mercurial-scm.org/wiki/Tag
- * This is only handling `:` in the same way as Git above
+(* Mercurial allows everything but :, \r and \n, but all these characters are
+ * unlikely to show up in versions so we just pass things through.
  *)
-let hg_escape_tag t = String.map (function ':' -> '_' | c -> c) t
+let hg_escape_tag t = t
 
-let hg_unescape_tag t = String.map (function '_' -> ':' | c -> c) t
+let hg_unescape_tag t = t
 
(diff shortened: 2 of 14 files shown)


NathanReb (commented):
This looks good, thanks!

I added a couple minor comments and have some more general extra notes:
- I think it'd be good to document that in the README. It's probably good to document it as part of `dune-release tag` documentation, also mentioning that people that don't use this command can use `_`. It's also probably worth adding a `Releasing a beta version` section that explains that `~` sorts first (e.g. that `1.0.0~beta1` is lower than `1.0.0~beta2` that is itself lower than `1.0.0`) that points to this feature as well. 
- We use a slightly different test convention than what you used. In the test file we group the test cases for the same function in a list that we bind to a variable called `test_<function_name>` and then concat all those lists into one because Alcotest don't allow arbitrary nesting. This results in the same list of test cases in the end but make it easier to modify the tests when functions are modified, renamed, removed etc... Would you mind following that same convention here? You can keep the factored `make_test` function you wrote and take inspiration from [this file](https://github.com/ocamllabs/dune-release/blob/main/tests/lib/test_github_repo.ml) for the grouping part!

Once we have that I think it's good to go!

lib/vcs.ml
```diff
@@ -189,34 +189,21 @@ let git_ls_remote ~dry_run r ~kind ~filter upstream =
...
-    |> Re.replace_string ~all:true unallowed_substrings ~by:"__"
+let re_tilde = Re.(compile @@ str tilde)
+
+let underscore = "_"
+
+let re_underscore = Re.(compile @@ str underscore)
+
+(* See the reference here: https://git-scm.com/docs/git-check-ref-format
+  * Similar escape as DEP-14: https://dep-team.pages.debian.net/deps/dep14/ *)
+let git_escape_tag t = Re.replace_string ~all:true re_tilde ~by:underscore t
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Yes, that's much better. I was looking for a straight `replace : search:string -> by:string -> string -> string` in `Astring` but it doesn't seem to exist. Fixed in a96d3b3.

CHANGES.md
```diff
@@ -8,6 +8,10 @@
 
 ### Fixed
 
+- Infer release version are inferred from tags, so users that don't use the
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Oh, good catch. Guess I got sloppy writing, especially since here the compiler won't help me. I rewrote that in different way altogether, d82292d.

lib/vcs.ml
```diff
@@ -134,7 +134,7 @@ let git_describe ~dirty r commit_ish =
   let dirty = dirty && commit_ish = "HEAD" in
   let git_describe =
     Cmd.(
-      git_work_tree r % "describe" % "--always"
+      git_work_tree r % "describe" % "--always" % "--tags"
```
NathanReb, line 137: Why is this needed? This allows including non annotated tags. That might be worth considering, it's been requested in the past, but I think it's unrelated to this bug fix.

lib/vcs.ml
```diff
@@ -433,9 +410,13 @@ let submodule_update ~dry_run r =
...
-let sanitize_tag = function
-  | `Git, _, _ -> git_sanitize_tag
-  | `Hg, _, _ -> hg_sanitize_tag
+let escape_tag = function
+  | `Git, _, _ -> git_escape_tag
+  | `Hg, _, _ -> hg_escape_tag
+
+let unescape_tag = function
+  | `Git, _, _ -> git_unescape_tag
+  | `Hg, _, _ -> fun x -> x
```
NathanReb, line 419: I think it's ok to remove the intermediate function but it's propably worth mentioning that `~` is a valid mercurial tag character then!

It would probably look better to have that comment in mercurial specific function but both work for me!

tests/lib/test_pkg.ml
```diff
@@ -95,22 +95,33 @@ let distrib_file =
...
-    make_test ~test_name:"3" ~opam:false ~version:"v0" "foo-v0";
-    make_test ~test_name:"4" ~opam:true ~version:"v0" "foo.v0";
-    make_test ~test_name:"5" ~opam:false ~tag:"v0" ~keep_v:false "foo-v0";
-    make_test ~test_name:"6" ~opam:true ~tag:"v0" ~keep_v:false "foo.0";
-    make_test ~test_name:"7" ~opam:false ~tag:"v0" ~keep_v:true "foo-v0";
-    make_test ~test_name:"8" ~opam:true ~tag:"v0" ~keep_v:true "foo.v0";
-    make_test ~test_name:"9" ~opam:false ~tag:"v0" ~version:"x" "foo-v0";
-    make_test ~test_name:"10" ~opam:true ~tag:"v0" ~version:"x" "foo.x";
+    make_test ~test_name:"from version while keeping v" ~version:"v0"
+      ~keep_v:true "foo.v0";
```
NathanReb, line 105: Do this tests still make sense since the version is used directly and therefore keep_v has no effect?

Wouldn't it be better to test how `Pkg.version` and `Pkg.tag` work instead, pontentially in a separate PR?

tests/lib/test_vcs.ml
```diff
@@ -1,25 +1,40 @@
...
-        name expected
-        (Vcs.git_sanitize_tag input)
+      Alcotest.(check Alcotest_ext.tag) name expected (Vcs.git_escape_tag input)
     in
     (name, `Quick, test_fun)
   in
   [
-    make_test "empty" ~input:"" ~expected:"";
-    make_test "valid" ~input:"3.3.4" ~expected:"3.3.4";
+    make_test "empty" ~input:"" ~expected:(t "");
```
NathanReb, line 19: The conversion can be factored into `make_test`!

NathanReb (commented):
Awesome! This looks great!

Do you think by any chance that the integration test could be trimmed down by grepping for a specific line in the output?

lib/vcs.ml
Leonidas-from-XIV: We have code where we parse unannotated tags and accept them while printing a warning, so I assumed it should be possible throughout. I can remove this but I feel like we should be consistent about unannotated tags.

NathanReb: It would probably have made sense to make a separate PR for this but let's not bother!

I think it's worth having its own changelog entry though!

lib/vcs.ml
Leonidas-from-XIV: I finally looked at what [Mercurial tags](https://www.mercurial-scm.org/wiki/Tag) support and decided to implement it properly. Or, well, close enough to be useful and basically analogous to Git.

tests/lib/test_pkg.ml
Leonidas-from-XIV: Honestly, I don't think they do. I had to remove some already, because there's no way to unit test these without running them in an existing git repo (so I converted them into an integration test to still keep the functionality tested) and the remaining ones are a bit nonsensical because they mostly test absolutely trivial things while looking like the expected cases are wrong (`~keep_v:false` yet the expected value has a `v` — I'm thinking that `keep_v` should be also renamed/refactored at a later point to something like `type version_transform = [`As_is, `Remove_v_from_tag]` but that's probably something for a future PR).

So I guess I'll just delete them and make a PR in the future to test this better.

NathanReb (approved):
Awesome, thanks! Love the concise tests!

PR #384, opened 2021-09-07
Refactor distrib tests

Leonidas-from-XIV:
This refactors the tests a little bit so they use Alcotest result testable. Also it gives them individual names.


gpetiot (approved):
Looks good to me

PR #386, opened 2021-09-09
Introduce Version.t and Vcs.Tag.t to be explicit

Leonidas-from-XIV:
Currently `dune-release` plays a bit fast and loose with what is a `version` and what is a `tag` and they are easy to accidentally swap. This PR aims to refactor the code to not accidentally convert between these.


bin/args.ml
```diff
@@ -0,0 +1,11 @@
+open Cmdliner
```
NathanReb, line 1: This can all be moved to the `Cli` module!

bin/check.ml
```diff
@@ -20,15 +20,18 @@ let check (`Package_names pkg_names) (`Package_version version) (`Dist_tag tag)
      if on_working_tree then (OS.Dir.current (), fun _ -> ())
      else
        let dir =
-         Pkg.tag_from_repo ?tag ?version () >>= fun inferred_tag ->
+         (* TODO this is a weird cludge because `Pkg.tag` wants a `Pkg.t` *)
```
NathanReb, line 23: I think the check command should likely create its own `Pkg.t` based on the CLI arguments, just as all other commands do.

The problem is that the `Pkg.t` type isn't super well defined and it in facts represent an abstraction of the repository rather than a single package. This is something we'll work on improving shortly for 2.0 but for now I think it's okay to just properly initialize one and pass it around same as in the other commands.

lib/github.ml
```diff
@@ -323,41 +323,45 @@ let assert_tag_exists ~dry_run tag =
    point to the same ref. *)
 let push_tag ~dry_run ~yes ~dev_repo vcs tag =
   let remote_has_tag_uptodate () =
```
NathanReb, line 325: How about we move this function to the toplevel of the module and also eventually extract other functions so its flow reads better? It was already quite convoluted and the triple level of nested pattern matching it has now didn't really make it any better.

lib/pkg.ml
```diff
@@ -204,20 +193,22 @@ let homepage_is_on_github p =
...
     match p.distrib_file with
     | Some f -> Fpath.basename f
-    | None -> "$(NAME)-$(TAG).tbz"
+    | None -> strf "%s-%a.tbz" name Vcs.Tag.pp tag
   in
-  dev_repo_is_on_github p >>= fun a ->
-  homepage_is_on_github p >>= fun b ->
-  let uri =
-    (if a || b then "releases/download/$(TAG)/" else "releases/") ^ basename
+
```
NathanReb, line 205: I'm surprised ocamlformat didn't get rid of that empty line but either I think we should drop it! I don't think it's intentional anyway, is it?

lib/vcs.ml
```diff
@@ -8,8 +8,27 @@ open Bos_setup
...
+module Tag = struct
+  type t = string
+
+  let pp = Fmt.string
+
+  let equal = String.equal
+
+  let to_string x = x
+
+  (* no escaping here in case the user wants to force a literal tag *)
```
NathanReb, line 20: We should probably make this a doc comment in the `.mli`!

NathanReb (commented):
It looks good, thanks! This is more than welcome!

I left a few comments but nothing major I think!

As we discussed offline, I think we should review how we use tags and versions, write it down in a document in the repo and eventually write a spec if we feel like something should be changed here!

gpetiot (commented):
It looks good to me, this part really needed a cleanup!

lib/version.mli
```diff
@@ -0,0 +1,11 @@
+type t
+
+val from_tag : keep_v:bool -> Vcs.Tag.t -> t
+
+val to_tag : t -> Vcs.Tag.t
+
+val from_string : string -> t
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I don't know. In dune-release it is about 50:50, we define `of_string` in `lib/opam.ml` but `from_string` in `json.ml` (probably because Yojson uses `from_string`), so maybe there is some value in consistency - whatever that consistency will be in the end.

lib/vcs.ml
Leonidas-from-XIV: I added documentation for the whole module in the `.mli`.

bin/check.ml
Leonidas-from-XIV: Thanks. The whole logic in `check` is rather complex, with `Pkg.v` created in various places in seemingly hackish ways so I guess this makes sense until the whole logic is revisited.

tests/lib/test_pkg.ml
```diff
@@ -1,6 +1,13 @@
 open Rresult
 open Dune_release
 
+(* compatibility shim with OCaml 4.06 and 4.07 *)
+module Option = struct
+  include Option
+
+  let map f = function None -> None | Some x -> Some (f x)
```
gpetiot, line 8: We already have redefined `Option.map` in `Dune_release.Stdext`

Leonidas-from-XIV: A bit unusual to include a labelled version there but I'll take it :-)

lib/github.ml
Leonidas-from-XIV: I tried refactoring it a bit. In general the way it determines a tag is annotated is kinda weird and I feel that functionality should eventually live somewhere else but I also think that that's a battle for another time.

Leonidas-from-XIV:
Since this is touching quite some amount of external things to make sure this works I forked @NathanReb's dune-release-testing repo and used dune-release to make a release, and a PR: https://github.com/Leonidas-from-XIV/opam-repository/pull/1

It looks like it works, so I think it is ready to be merged when people are happy with the state of things.

lib/pkg.mli
```diff
@@ -154,7 +149,7 @@ val build : f
 
 (** {1 Version} *)
 
-val extract_version : t -> (string, Sos.error) result
+val extract_version : t -> (Version.t, Sos.error) result
```
NathanReb, line 152: Same as I mentioned above but this should probably return a string to avoid confusion. Using a `Version.t` is confusing as I wouldn't expect a `Version.t` to contain the leading `v` if `keep-v` is `false` for example.

NathanReb: Might be worth extracting a new type while we're at it. Might be a `Changelog_version.t` or `Version.Changelog.t`. That would also prevent us from badly converting from this type to `Tag.t` or `Version.t`. What do you think?

lib/github.ml
```diff
@@ -304,15 +304,65 @@ let check_tag ~dry_run vcs tag =
...
+  let missing_tag pp_r =
+    App_log.unhappy (fun l ->
+        l
+          "The tag %a is present on the remote but points to a different \
+           commit (%a)."
+          Text.Pp.tag tag pp_r ())
+  in
+  match Vcs.commit_id ~commit_ish:remote_rev_unpeeled vcs with
+  | Ok remote_rev
+    when remote_rev = local_rev && is_annotated_tag vcs remote_rev_unpeeled ->
```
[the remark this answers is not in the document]
Leonidas-from-XIV: The reason why it is calling `Vcs.commit_id` is because for extracting a function like `is_annotated_tag` I think it makes sense that the function would get the "unpeeled" commit, resolve it, and see itself. Yet here it calls `Vcs.commit_id` to see if there is a commit to begin with. While it is a bit ugly that it ends up calling `Vcs.commit_id` effectively twice I think it makes the code more understandable than before since the intent is now clearer.

The alternative is basically rewriting the function to

```ocaml
let is_annotated_tag tag resolved_tag = String.equal tag annotated_tag
```
(excuse the ETA-expansion for clarity)

which would work but the function would be somewhat nonsensical because it doesn't determine anything but rather requires that the candidate annotated and the resolved tag are passed to it.

lib/github.ml
```diff
@@ -304,15 +304,65 @@ let check_tag ~dry_run vcs tag =
...
-      tag
+      Vcs.Tag.pp tag
 
 let assert_tag_exists ~dry_run tag =
   Vcs.get () >>= fun repo ->
   if Vcs.tag_exists ~dry_run repo tag then Ok ()
-  else R.error_msgf "%s is not a valid tag" tag
+  else R.error_msgf "%a is not a valid tag" Vcs.Tag.pp tag
+
+let is_annotated_tag vcs remote_rev =
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Personally I would like to come up with a better name for "unpeeled" in general because to me it is quite confusing to have negation in the name (I don't think there is a git term for it) so maybe "wrapped" or "outer" would make sense? It is even more annoying because at this point we don't know whether it is an "unpeeled"/"wrapped"/"outer" commit at all, that's exactly what we're trying to determine in this function. I would be very glad for a good name.

(But `remote_rev` is also wrong. I think in this function it should be called `commit_ish`, because the function is more general and checks exactly for that, even despite only being passed tags)

lib/github.ml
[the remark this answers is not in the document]
NathanReb: Ok I get it now!

I think a good alternative is to extract a function that will get the `remote_rev` and the `local_rev`, do the local de-referencing and return a value that carries all the information, something like ``[`Up_to_date_annotated | `Up_to_date_non_annotated | `Points_to_different_commit | `Points_to_missing_object]``.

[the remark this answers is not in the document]
Leonidas-from-XIV: That's a great idea. Also rerenamed the functions to be closer to the original name and less vague.

lib/github.ml
```diff
@@ -304,15 +304,67 @@ let check_tag ~dry_run vcs tag =
...
+
+(* Resolve again in case of annotated tags (most common case).
+   This is a no-op for non-annotated tags. In case of error, we
+   can assume that the remote is different because we checked that we
+   have the tag locally. *)
+let determine_remote_tag_status ~local_rev ~remote_rev vcs =
+  match Vcs.commit_id ~commit_ish:remote_rev vcs with
+  | Ok resolved_rev when resolved_rev = local_rev && resolved_rev = remote_rev
+    ->
+      (* the resolved_rev was the same as remote, thus it is unannotated *)
```
NathanReb, line 325: Isn't it the other way around?

If it's annotated, there's an intermediate object. `ls-remote` will return this object's hash which we need to resolve again locally to obtain the pointed commit hash.

lib/pkg.ml
```diff
@@ -70,21 +62,25 @@ let change_log p =
...
+let version_from_tag pkg = tag pkg >>| Version.from_tag ~keep_v:pkg.keep_v
+
 let extract_version pkg = change_log pkg >>= fun cl -> extract_version cl
 
 let infer_version pkg =
-  match extract_version pkg with Ok t -> Ok t | Error _ -> tag pkg
+  match extract_version pkg with
+  | Ok changelog_version ->
+      Ok (Version.Changelog.to_version ~keep_v:true changelog_version)
```
NathanReb, line 75: I think we should use `Pkg.keep_v` here!

lib/github.ml
Leonidas-from-XIV: No, I think this is correct. If the `resolved_rev` is the same as the `remote_rev` that means they were pointing to the same commit, that is, there was no intermediate object.

Looking at the original source code on `main`:

https://github.com/ocamllabs/dune-release/blob/f3f38cc463eaa90d44661f8fa41c8eccc55ea2c7/lib/github.ml#L329-L341

`remote_rev_unpeeled` is what is `remote_rev` here and `remote_rev` is what this branch calls `resolved_rev`. The logic is indeed not great, so for the future I would suggest creating something like `Vcs.is_annotated_tag` but for I think this refactoring makes it at least better than what was before.

NathanReb (approved):

NathanReb:
Let's merge! Thanks for working on this!

PR #388, opened 2021-09-09
Fix `---V` to be `-V`

Leonidas-from-XIV:
Cmdliner accepts everything that is longer than one character as a long option, hence it made `-V` into `---V` long option.

This has been broken since `-V` was added in https://github.com/ocamllabs/dune-release/commit/809f5391dad64c32d381c550dc56ec4233336b1a, so maybe nobody uses it and it might sense to just remove that option?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Change the `---V` command option to be `-V` (#388, @Leonidas-from-XIV)
+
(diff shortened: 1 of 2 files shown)


NathanReb (approved):

NathanReb:
The `--pkg-version ...` option worked all along so I think we can keep it around. It's just the "short" option that was broken.

PR #389, opened 2021-09-10
Check for the tag to be a tag, not just a valid ref

Leonidas-from-XIV:
The code just checked the ref to exist which checked whether the ref existed. When passing in an (existing) commit hash this would return true, because a commit hash is a valid ref, but it is not a tag. This changes the code so it actually tests that the passed reference is indeed a tag.

I ran across this as I was accidentally passing commithashes to it (since the tagging failed) and it still accepted them.

This is not a user visible change, so I'm not adding this to the changelog, this is purely about correctness of the code.

### DIFF
--- a/tests/bin/non-github-doc-uri/run.t
+++ b/tests/bin/non-github-doc-uri/run.t
@@ -45,10 +45,12 @@ We do the whole dune-release process
 
 (1) distrib
 
-    $ dune-release distrib --dry-run
+    $ dune-release distrib --dry-run | grep -vE "Commit [a-f0-9]{40}"
+    dune-release: [ERROR] Github development repository URL could not be
+                          inferred.
     [-] Building source archive
     => rmdir _build/whatever-0.1.0.build
-    -: exec: git --git-dir .git rev-parse --verify 0.1.0
+    -: exec: git --git-dir .git rev-parse --verify refs/tags/0.1.0
     => exec: git --git-dir .git show -s --format=%ct 0.1.0^0
     => exec: git --git-dir .git clone --local .git _build/whatever-0.1.0.build
     => exec:
@@ -79,8 +81,6 @@ We do the whole dune-release process
     [ OK ] lint opam file whatever.opam.
     [ OK ] opam field synopsis is present
     [FAIL] opam fields homepage and dev-repo can be parsed by dune-release
-    dune-release: [ERROR] Github development repository URL could not be
-                          inferred.
     [FAIL] opam field doc cannot be parsed by dune-release
     [FAIL] lint of _build/whatever-0.1.0 and package whatever failure: 1 errors.
     
@@ -95,9 +95,7 @@ We do the whole dune-release process
     [ OK ] package(s) pass the tests
     
     [+] Distribution for whatever 0.1.0
-    [+] Commit ...
     [+] Archive _build/whatever-0.1.0.tbz
-    [1]
 
 (2) publish doc
 
--- a/tests/bin/non-github-doc-uri/run.t
+++ b/tests/bin/non-github-doc-uri/run.t
@@ -45,10 +45,9 @@ We do the whole dune-release process
 
 (1) distrib
 
-    $ dune-release distrib --dry-run | grep -vE "Commit [a-f0-9]{40}"
+    $ dune-release distrib --dry-run | grep -vE "Commit [a-f0-9]{40}" | grep -v "Building source archive"
     dune-release: [ERROR] Github development repository URL could not be
                           inferred.
-    [-] Building source archive
     => rmdir _build/whatever-0.1.0.build
     -: exec: git --git-dir .git rev-parse --verify refs/tags/0.1.0
     => exec: git --git-dir .git show -s --format=%ct 0.1.0^0
--- a/lib/vcs.ml
+++ b/lib/vcs.ml
@@ -141,7 +141,8 @@ let git_describe ~dirty r commit_ish =
   run_git_string ~dry_run:false r git_describe ~default:Default.string
 
 let git_tag_exists ~dry_run r tag =
-  match run_git_quiet ~dry_run r Cmd.(v "rev-parse" % "--verify" % tag) with
+  let tag_rev = "refs/tags/" ^ tag in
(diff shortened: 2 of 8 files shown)


NathanReb (approved):
This looks good, thanks!

I just noticed we redefine a `assert_tag_exist` function in three different modules. Might be good to factorize all this and move that function to `Vcs` at some point as well!

lib/pkg.ml
```diff
@@ -465,6 +464,7 @@ let distrib_archive ~dry_run ~keep_dir ~include_submodules p =
   else Ok ())
   >>= fun () ->
   distrib_prepare ~dry_run ~dist_build_dir ~version >>= fun () ->
+  assert_tag_exists ~dry_run repo_vcs tag >>= fun () ->
```
NathanReb, line 467: Why did you move this line around?

tests/bin/non-github-doc-uri/run.t
```diff
@@ -95,9 +95,7 @@ We do the whole dune-release process
     [ OK ] package(s) pass the tests
     
     [+] Distribution for whatever 0.1.0
-    [+] Commit ...
     [+] Archive _build/whatever-0.1.0.tbz
-    [1]
```
NathanReb, line 100: Is it intentional to remove this line? Those tests need serious rework but until then, I think the exit status actually is valuable here!

NathanReb (commented):
Something isn't clear to me here. It appears that this actually breaks some tests. It might be fine but we need to clearly identify what it is, how to fix the tests and add a changelog entry if the behaviour does change.

My guess is adding a sensible `git tag` command to some of the test that used to work without a tag might be enough

lib/pkg.ml
Leonidas-from-XIV: The problem is that the test checks for `assert_tag_exists` before it runs `dune subst`.

Before the test it always succeeded because `assert_tag_exists` only tested whether whatever commit it got is a commit (basically a no-op) and then it ran `dune subst` which failed with the error message that the tests expect.

After fixing the problem arises that `assert_tag_exists` actually *fails* here, because there is no tag (given this is `check` it rightfully doesn't create any tag), so it exits with the error message that the tag is missing instead.

Hence I moved it around. I am not particularly happy about it because the function does everything in a sort of random order and the expect tests are very strongly coupled to which order it does things in.

tests/bin/non-github-doc-uri/run.t
Leonidas-from-XIV: You're right, I have to investigate why the exit code changes suddenly. That's most likely wrong, given the package has linting errors.

Leonidas-from-XIV: So after I looked into this it is because now the error state reported is of `grep` and not `dune-release`. I've committed a fix but it only works on systems that use Bash as `$PIPESTATUS` is a bashism. Checking CI that means it works on Fedora, CentOS, OpenSUSE but fails on Debian, Ubuntu and Alpine. I'm not particularly fond of this solution so maybe we should just not test the error code in this case.

Leonidas-from-XIV:
I've changed the failing test to `tag` before `distrib`. This technically is a change in behaviour, since before it was possible to create untagged tarballs (since nothing was checking for the existence of the tag), but I believe the *intent* of the code was to enforce exactly the behaviour here.

Regular users of `dune-release` should not be affected since people will run `dune-release tag` before or just tag themselves, but I feel it is important to point that out.

tests/bin/non-github-doc-uri/run.t
NathanReb: I think the exit status ins't that important here and checking it prints the error message we're after should be enough.
This test could really use a description btw, it's really hard to understand what we're trying to test.

I assume we want to see the error messages for github URLs parsing failure, I guess all the rest is pretty much worthless.

Could you have a look at the original PR see what this is supposed to test so we can polish it a bit (in a separate PR)?

lib/pkg.ml
NathanReb: We should really avoid making changes like this just to get the tests to pass.

Which test was failing? From what I understood it's related to the check command and should work without a tag (which actually sounds good). I think we should look into this, eventually exposing an intermediate function that builds a distrib archive for any commit-ish (used by check) and one above it that runs all the proper checks (used by distrib).

Then we wouldn't have to change that code, and `check` would do the right thing. How does that sound?

Leonidas-from-XIV: The test that fails is `tests/bin/distrib-name` where it uses `dune-release distrib` to check that `dune-release` will fail when `dune subst` can't find the name of a project. Without that change `dune subst` is never called because it fails earlier that the tag does not exist.

https://github.com/ocamllabs/dune-release/blob/f3f38cc463eaa90d44661f8fa41c8eccc55ea2c7/tests/bin/distrib-name/run.t#L56-L64

What the code on `main` does *is* the right behaviour — Calling `dune subst` will `git describe --always --dirty` which will write basically something somewhat random into the files, not the tag that we want.

Upon writing this I realized I was barking up the wrong tree. The code had the right behaviour and what I need to do is to fix the test to actually tag and then it all makes sense again.

Your suggestion to make an intermediate function is a good one though.

tests/bin/non-github-doc-uri/run.t
Leonidas-from-XIV: I think the original PR is #256 and I think you're right, the only thing that is important is the error message. I would suggest at some point going through all the tests, identifying what is important and then grepping for that exact thing.

NathanReb:
> I've changed the failing test to tag before distrib. This technically is a change in behaviour, since before it was possible to create untagged tarballs (since nothing was checking for the existence of the tag), but I believe the intent of the code was to enforce exactly the behaviour here.

It is one indeed and it is well worth pointing it out. That being said, I don't think it's too critical as if you tried releasing without a tag, dune-release would fail at the subsequent step when trying to create a Github release. Even for non github users, I'm not sure we should allow releasing without a tag.

If this turns out to be bothering a lot of users, it can be properly fixed by simply using the output of `git describe` to create the tarball while we keep using the tag everywhere else. I doubt it'll come to this though.

PR #390, opened 2021-09-17
Consistent error reporting

Leonidas-from-XIV:
Experiment on whether we can replace `Logs.on_error_msg` with something that uses our own logging functions and is more consistent. We had some ordering issues so this should help.

### DIFF
--- a/tests/bin/non-github-doc-uri/run.t
+++ b/tests/bin/non-github-doc-uri/run.t
@@ -45,9 +45,7 @@ We do the whole dune-release process
 
 (1) distrib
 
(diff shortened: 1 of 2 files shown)


NathanReb (approved):
So you managed to just fix it in the test in the end? Cool!

Leonidas-from-XIV:
Yes, I realized that the error message is emitted on `stderr` which is unaffected by the `| grep` thus depending on I guess the interplay of shell, `grep` and libc the way they are output can be different. Piping all of it into `stdout` simplifies that a bit because it is at least the same channel and gets processed by `grep`.

PR #392, opened 2021-09-27
No more bistro

### LINKED ISSUE #241: Remove the bistro command
Calling `dune-release` without a sub-command is the same as calling `dune-release bistro`. Is there really a benefit to having this command? (not sure this is even properly documented)

Leonidas-from-XIV:
Since we're preparing 2.0.0, might as well remove the `bistro` command since it is equivalent to just calling `dune-release`.

Closes #241.


NathanReb:
This is good but we should release the latest bug fixes first!

Leonidas-from-XIV:
@NathanReb Looks like it is time to tackle this, since we got all the bugfixes released. I've rebased on the current `main`.

gpetiot (approved):
Looks good to me!

PR #393, opened 2021-09-30
Revert use-yaml branch merge

Leonidas-from-XIV:
We concluded that YAML support is not something that we want to have since nobody actually likes the YAML format and creating a YAML parser is sufficiently difficult that the only good library for it uses ctypes to bind to libyaml. A more realistic alternative is switching the configuration to use s-expressions and use a proper s-expression parser in the medium term since we will need this to parse `dune-project` files anyway and this way we avoid the proliferation of various ad-hoc config files in the OCaml ecosystem.

This reverts commit 901a95402dc534b1cd067d51b68ffb1f64406aa6, reversing changes made to cee97c1235aadacd24474d798f4ba2ee6e1a4d0e.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -39,5 +39,4 @@ with [Dune](https://github.com/ocaml/dune) and hosted on
   odoc
   (alcotest :with-test)
   (mdx (and :with-test (>= 1.6.0)))
-  yaml
   yojson))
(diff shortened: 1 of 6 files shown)


NathanReb (approved):
:+1:

PR #400, opened 2021-10-12
Port `mdx` tests to `cram` tests

Leonidas-from-XIV:
This is a mostly mechanical translation of the tests to use `cram` instead of `mdx`. The main pain points were indeed the lack of `...` but on the other hand we gain mostly robust support for `dune promote`.

I added an additional helper, `make_dune_release_deterministic` that does a bit more normalization where it was needed after fighting for a while with `sed` & `grep`. Maybe it would make sense to merge it with `make_check_deterministic` into a general `make_deterministic` binary that can be used by all tests, but for now these are separate.


tests/bin/distrib-name/run.t
```diff
@@ -4,82 +4,82 @@ dune-project. This goes with the fix for #320, where:
...
+  $ git add liba/* libb*/ CHANGES.md README LICENSE *.opam dune-project .gitignore
+  $ git commit -m 'Commit.' > /dev/null
 
 Try dune-release distrib with no project name.
 
-    $ dune-release distrib --skip-lint
-    [-] Building source archive
-    dune-release: [ERROR] cannot determine distribution name automatically: add (name <name>) to dune-project
-    [1]
+  $ dune-release distrib --skip-lint
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Even better, since the error messages are output to `stderr` we can just pipe `stdout` to `/dev/null` and get all the relevant errors that we're looking for.

Leonidas-from-XIV:
I've gone through the tests one by one, silencing uninteresting output, grepping for relevant lines and adding descriptions of what the test is looking for. Also joined the helper binary into a single executable that does all the relevant replacements where going manually with Unix commands would've been too painful.

tests/bin/distrib-name/run.t
```diff
@@ -1,85 +1,80 @@
...
-    [-] Running package tests in ...
-    [ OK ] package(s) pass the tests
-    
-    [+] Distribution for liba ...
-    [+] Commit ...
-    [+] Archive ...
+  $ git add dune-project && git commit -m 'add name' > /dev/null
+  $ dune-release distrib --skip-lint | make_dune_release_deterministic
+  [-] Building source archive
+  [+] Wrote archive _build/PKG(liba)-VER(0.42.0)-SINCE(1)-COMMIT(<deterministic>).tbz
```
NathanReb, line 70: I'm a bit sceptical about this rewriting, it's a bit confusing. I wonder if it wouldn't be best to live the deterministic parts untouched and rewrites the bits that are with something a bit less invasive.

What's the original output we're rewriting here?

tests/bin/draft/run.t
```diff
@@ -1,78 +1,51 @@
...
 
-We do the whole dune-release process
+We do the whole `dune-release` process
 
 (1) distrib
 
-    $ dune-release distrib --dry-run > /dev/null
+  $ dune-release distrib --dry-run > /dev/null
 
 (2) publish
```
NathanReb, line 45: Might be worth a word about what that should do, i.e. that it should create a draft release on the GH repo, setting draft to true in the API request, writing the release ID in a `.draft_release` file for later consumption by undraft.

NathanReb: I'm a bit surprised this test doesn't run `opam submit` in dry-run mode but that's a matter for a different PR!

tests/bin/errors/run.t
```diff
@@ -1,54 +1,54 @@
...
-    [-] Using tag "0.1.0"
-    dune-release: [ERROR] Due to invalid commit-ish `1`:
-      Exit code 128 from command
-        `git --git-dir .git rev-parse --verify 1^0`:
-      fatal: Needed a single revision
+  $ dune-release tag --commit=1
+  [-] Extracting tag from first entry in CHANGES.md
+  [-] Using tag "0.1.0"
+  dune-release: [ERROR] Due to invalid commit-ish `1`:
+    Exit code 128 from command
```
NathanReb, line 50: Weird trailing whitespace!

tests/bin/helpers/make_dune_release_deterministic.ml
```diff
@@ -0,0 +1,88 @@
...
+let make_test_directory_deterministic line =
+  (* OCaml implementation of bash's
+     sed 's/\/.*\/tests\/bin\/check/<test_directory>/' *)
+  let re = compile @@ seq [ str "/"; rep any; str "/tests/bin/check" ] in
+  replace_string re ~by:"<test_directory>" line
+
+let () =
+  try
+    while true do
+      () |> read_line |> make_build_deterministic |> make_commit_deterministic
```
NathanReb, line 84: Probably worth having a list of transformation and applying them via a `List.fold`, what do you think?

tests/bin/include-submodules/run.t
```diff
@@ -1,45 +1,35 @@
...
-    $ dune-release distrib --skip-lint --skip-build --skip-test --include-submodule --dry-run
-    ...
-    => chdir _build/whatever-0.1.0.build
-       [in _build/whatever-0.1.0.build]
-    -: exec: git --git-dir .git submodule update --init
-    ...
-    [+] Wrote archive _build/whatever-0.1.0.tbz
-    ...
-    [+] Archive _build/whatever-0.1.0.tbz
+  $ dune-release distrib --skip-lint --skip-build --skip-test --include-submodule --dry-run | grep -- "--init"
```
NathanReb, line 34: Any particular reason to grep for this rather than `git submodule update`?

NathanReb (commented):
It looks good!

I have few comments but overall it looks great and we will benefit from this change even if we can further improve the tests in future PRs!

tests/bin/include-submodules/run.t
Leonidas-from-XIV: Yes, with git it is quite easy to add `git [options-here] submodule [maybe-options-here?] update [definintely-options-here]` so I felt that that would've been somewhat brittle, whereas `--init` is unlikely to be removed from the git call. And given the command will be printed it will still check that it is `git submodule update` and not some other `git` subcommand that uses `--init`.

In practice I could also `grep` for `submodule` or `update` but these are more likely to be printed by `dune-release` itself. Which is not a big issue since an additional line isn't a problem, just `--init` seemed to me like the safest bet.

tests/bin/distrib-name/run.t
Leonidas-from-XIV: You're right. Looking back I don't know why I did this wrapping. Maybe because I had fun with the `Re` combinators but of course that rewriting doesn't add anything.

The original line goes something like `_build/liba-0.42.0-1-gbadf00d`. I'll change the helper to just leave them be since everything but the commithash is reproducible anyway.

tests/bin/helpers/make_dune_release_deterministic.ml
Leonidas-from-XIV: That's a good idea, especially as `ocamlformat` doesn't format this in a nice 1-transform-per-line way.

tests/bin/include-submodules/run.t
NathanReb: Sounds good to me, let's keep `--init`!

tests/bin/distrib-name/run.t
NathanReb: Looks great now!

NathanReb (approved):
Looks good, thanks!

PR #404, opened 2021-10-28
Make all packages use the same asset url file

### LINKED ISSUE #402: with 1.5.1, bad urls for subsequent packages
I ran dune-release distrib ... dune-release opam pkg / dune-release opam submit, and only the "main" package got a "v" in the url, the others did not (leading to 404 from opam-ci): https://github.com/ocaml/opam-repository/pull/19818 fix in https://github.com/ocaml/opam-repository/pull/19818/commits/9d5053d20c11f1dcb78abd8d31b9b6da9edced73

not sure why this happens, and if this is an issue with happy-eyeballs (see https://github.com/ocurrent/ocaml-ci/issues/381 which causes some other trouble). a local `dune build` does not work, but `dune build -p happy-eyeballs` and `dune build -p happy-eyeballs-lwt,happy-eyeballs-mirage` do.

Leonidas-from-XIV:
Fixes #402


NathanReb (commented):
I wonder if a better scoped fix wouldn't be to simply change the url file name, not the archive name.

We'd still build the same archive but look for the url in something like `version.url` or `tag.url` or whatever name is tied to the version we're releasing. Would that work?

NathanReb:
Because if we're to change the tarball name, we might as well start naming it after the repository directly rather than going through this `asset-version` intermediate phase.

Leonidas-from-XIV:
> I wonder if a better scoped fix wouldn't be to simply change the url file name, not the archive name.

This is exactly what it is doing. It changes from `$projectname-$version.url` to `asset-$version.url`. I kept the version to keep some entropy and just replaced the `$projectname` with the fixed string `asset`. The reason was that I wanted to keep the change small in the context of a bugfix.

Overall I agree that in the medium run we should name it after `$reponame-$version.url` but I believe that is a somewhat more controversial change that's visible from outside the dune-release "internals" so while I'll happily do that (and would solve #366 as well) I'd suggest this as a side-step to be able to release a 1.5.2 soon.

NathanReb (approved):
My bad! I misinterpreted some of the changes in the tests, thought the name of the distribution tarball changed as well.

Looks good then! Thanks for looking into this!

PR #411, opened 2021-11-02
Remove support for OPAM 1.2.2 client

### LINKED ISSUE #406: Remove support for opam 1.2 CLI
At the moment we still support the opam 1.2.2 client. There are [very few distributions](https://repology.org/project/opam/versions) that do not have an opam 2.x available these days (mostly Gentoo forks like Exherbo and Funtoo), so this support these days is mostly untested (we don't test it in CI because we don't have an opam 1.2 system available), OPAM Repository doesn't continue to maintain 1.2 format packages.

For this reason it would be reasonable to remove support for the old version to simplify our code.

Users that are stuck on very old Debian or Ubuntu versions can continue to use the pre-2.x versions of `dune-release` which continue to be available, or using a third-party repository to get a newer OPAM. Given releases are usually done from developer machines the impact of not supporting 2017-era distributions should be minimal.

Leonidas-from-XIV:
Closes #406


NathanReb (approved):
Looks good, thanks!

I was looking at the opam lint code while reviewing this PR. It mentiones opam lint CLI is "broken". We should maybe report this upstream so we can simplify our opam lint invocation in the future!

Leonidas-from-XIV:
@pitag-ha @NathanReb We should try to call opam lint via the API but in the meantime I think it is a good step to merge this to prevent the branch from bitrotting.

PR #412, opened 2021-11-05
Add dependency to `opam-file-format`

Leonidas-from-XIV:
The module `OpamParserTypes.FullPos` is from ocaml-file-format and if too low of a version is selected ([like `2.0.0~rc2` by the OCaml CI](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/b6acd886778d49581c876692880f381ddd813cd0/variant/opam-2.1,compilers,4.06,dune-release.1.5.2,lower-bounds)) as a dependency of `opam-format` then the build will fail.

Therefore we need `opam-file-format` as direct dependency.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -33,9 +33,9 @@ with [Dune](https://github.com/ocaml/dune) and hosted on
   (re (>= 1.7.2))
   astring
   (opam-file-format (>= 2.1.2))
-  (opam-format (>= 2.1.0~beta))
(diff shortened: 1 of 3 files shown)


NathanReb (approved):
Looks good to me!

We might want to bump the required versions for the other opam libraries to `>= "2.1.0"`, no need to support the beta versions anymore I'd say!

PR #413, opened 2021-11-05
Specify a message in the git tests

Leonidas-from-XIV:
The ancient git version in CentOS 7 (1.8.3.1) has a but where despite `--allow-empty-message` it will not accept `-m ''` so to fix this we set a (boring) message to commit.


Leonidas-from-XIV:
Given we don't test on CentOS 7 and it ships an ancient git I am not entirely convinced myself supporting it is worth it, because who knows what else is broken.

NathanReb (approved):

PR #415, opened 2021-11-16
tests: ensure user.email and user.name is always set to allow commiting

sternenseemann:
Similar regression as also fixed in #334: We need to ensure that the git user's name and email are set, otherwise we can't create a commit if no global git configuration has been set up.


NathanReb (approved):
Looks good, thanks for fixing this!

I guess ocaml-ci must have those set since the CI didn't catch this!

PR #421, opened 2021-12-26
Parse change logs in the style of keepachangelog.com

ifazk:
I recently started using the change log format specified in https://keepachangelog.com. This format uses "Changelog" as the first level of header.

This resulted in `dune-release` trying to use `"Changelog"` as the tag. I manually specified the tag, but then `dune-release` used the entire changelog for the pull-request message, see https://github.com/ocaml/opam-repository/pull/20195.

This PR changes `Text.change_log_last_entry` to recurse into `changes` if the header title is "Changelog", and drops brackets around version numbers if there are any.

Btw, I wasn't sure if PRs from outsiders are welcome or not and I didn't find a contribution guide. Hope this doesn't feel too aggressive since I didn't open an issue first.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,15 +4,14 @@
 
 - Add `--skip-lint`, `--skip-build`, `--skip-test` and
   `--keep-build-dir` to the main command (#419, @NathanReb)
+- Added support for parsing changelogs written in the style of
+  [keepachangelog.com](https://keepachangelog.com/) (#421, @ifazk)
 
 ### Changed
 
 ### Deprecated
 
 ### Fixed
-
-- Fixed parsing the changelog if it is written in the style of
-  keepachangelog.com(https://keepachangelog.com/) (#421, @ifazk)
 
 ### Removed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@
 ### Fixed
 
 - Fixed parsing the changelog if it is written in the style of
-  keepachangelog.com(https://keepachangelog.com/) (#<PR_NUMBER>, @ifazk)
+  keepachangelog.com(https://keepachangelog.com/) (#421, @ifazk)
 
 ### Removed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@
 
 ### Fixed
 
+- Fixed parsing the changelog if it is written in the style of
+  keepachangelog.com(https://keepachangelog.com/) (#<PR_NUMBER>, @ifazk)
+
 ### Removed
 
 ### Security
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -11,6 +11,9 @@
 
 ### Fixed
 
+- Fixed parsing the changelog if it is written in the style of
+  keepachangelog.com(https://keepachangelog.com/) (#421, @ifazk)
```
Leonidas-from-XIV, line 15: Could you move this to the "added" section? I don't think it is a fix, because the fact that it didn't use some specific format wasn't "broken", it was just unsupported. Now you added support :-)

The markdown syntax for the link is wrong, it needs `[keepachangelog.com](https://keepachangelog.com/)`

lib/text.ml
```diff
@@ -100,26 +100,39 @@ let header_title ?(flavour = `Markdown) h =
...
+              | "" :: lines -> String.concat ~sep:"\n" lines
+              | lines -> String.concat ~sep:"\n" lines
+            in
+            let version =
+              let fst = String.get version 0 in
+              let len = String.length version in
+              let last = String.(get version (len - 1)) in
+              if List.mem fst [ '('; '[' ] && List.mem last [ ')'; ']' ] then
+                (* Drop delimitors for keepachangelog.com style *)
+                String.with_index_range ~first:1 ~last:(len - 2) version
```
Leonidas-from-XIV, line 132: I am not a big fan of index magic because it is tricky to read and a lot of code to begin with, only to realize that what it is doing is basically just trimming the delimiters.

Knowing that, I wondered how this could be rewritten in a more descriptive way since complaining without constructive critique is a bit silly and came up with this:

```ocaml
let version_delimiters = function '(' | ')' | '[' | ']' -> true | _ -> false in
let version = String.trim ~drop:version_delimiters version
```
This does the same thing but I think it is easier to read and understand. It also gets rid of the conditional because if there are no delimiters it is idempotent.

(One could argue that it accepts and removes the delimiters from `[0.0.1)` which is technically probably wrong but pragmatically this is probably good enough)

Leonidas-from-XIV (changes requested):
Hi @ifazk, thanks for your PR. Contributions are most definitely welcome! I thought this was somewhat obvious but it is a good point and we should probably create a `CONTRIBUTING.md` to describe what and how.

As to the PR itself, I wasn't aware of this format before but the changes required to also support that are rather small so I think this is a fine addition. I appreciate that you added tests for it, because otherwise it would be prone to breaking accidentally. I wrote some comments for improving it, so if you could address these I think it is ok to merge.

lib/text.ml
ifazk: Wasn't really happy with that part of the code myself. I've implemented your version.

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
Our CI formatting linter is a bit broken, so I'm merging after checking it locally. Thanks!

ifazk:
Thanks for your help and guidance in getting this merged!

PR #425, opened 2022-01-07
Add `CONTRIBUTING.md`

### LINKED ISSUE #423: Create `CONTRIBUTING.md`
In #421 a good point was raised that we don't have a stated policy on contributions. I am sure we can take some inspiration from MDX' [`CONTRIBUTING.md`](https://github.com/realworldocaml/mdx/blob/main/CONTRIBUTING.md) and add it to this project so people know that outside contributions are definitely welcome!

Leonidas-from-XIV:
I took the quite comprehensive file from MDX and applied it to `dune-release`. I am sure more can be written but maybe this is an ok place to start?

Closes #423

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,11 @@
+## General
+
+While this project is in the OCaml Labs organization, contributions from others
+are very much welcome. We're trying to build a tool that works for many users.
+
+If you're unclear whether a feature would fit the scope of `dune-release` don't
+hesitate to open an issue to gauge the interest.
+
 ## Setting up your working environment
 
 If you want to contribute to the project you'll first need to install the dependencies.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,10 +6,14 @@ You can do it via `opam`:
 ```sh
 $ git clone git@github.com:ocamllabs/dune-release.git
 $ cd dune-release
+$ opam switch create ./ ocaml-base-compiler.4.13.1 --no-install
 $ opam install -t --deps-only .
 ```
 
-This will install both regular and test dependencies to your active opam switch.
+This will create a local switch with a fresh compiler, the dependencies of
+`dune-release` and the dependencies for running the tests. The exact OCaml
+version is just an example, you should be able to use any reasonably recent
+version of OCaml.
 
 From there you can build `dune-release` by simply running:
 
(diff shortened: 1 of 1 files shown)


ifazk:
Could we add a statement at the beginning saying "Contributions are welcome" or something similar?

CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
+## Setting up your working environment
+
+If you want to contribute to the project you'll first need to install the dependencies.
+You can do it via `opam`:
+
+```sh
+$ git clone git@github.com:ocamllabs/dune-release.git
+$ cd dune-release
+$ opam install -t --deps-only .
```
NathanReb, line 9: We might want to replace that with the proper `opam switch create ./` invocation, what do you think?

NathanReb (commented):
This looks great! Thanks for taking the time to add this!

I'm happy with an extra comment mentioning contributions are welcome as suggested by @ifazk, if this encourage external contributors to get involved it's for the best!

Only a minor comment about setting up dependencies as I think we should encourage people to use local switches!

Leonidas-from-XIV:
@NathanReb @ifazk I've incorporated your suggestions, does that sound good to you?

CONTRIBUTING.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Setting up your working environment
+
+If you want to contribute to the project you'll first need to install the dependencies.
+You can do it via `opam`:
+
+```sh
+$ git clone git@github.com:ocamllabs/dune-release.git
+$ cd dune-release
+$ opam switch create ./ ocaml-base-compiler.4.13.1 --no-install
```
NathanReb, line 17: I think this can be abreviated `opam switch create ./ --deps-only -t`, let me check !

Leonidas-from-XIV: That's a good point, I wasn't aware this is possible. I had to use `opam switch create ./ ocaml-base-compiler.4.13.1 --deps-only -t`, otherwise the switch wouldn't have a compiler invariant and that seemed somewhat bad to me.

NathanReb: I think it's alright not to have an invariant here but we can always change that later!

ifazk:
> @ifazk I've incorporated your suggestions, does that sound good to you?

I'm really happy with the additions! Thanks!

Leonidas-from-XIV:
Yup, let's merge. If we want to revise we can do so in a later PR too.

PR #428, opened 2022-02-10
Remove delegates

### LINKED ISSUE #188: Replacing delegates

NathanReb:
Closes #188

It's finally here! This is also breaking so we should release whatever we already have if we want them available under 1.x.


NathanReb:
This was pretty simple and straightforward, I think most of the work will be improving a bit how we detect whether the packages are hosted on github and where. This still uses homepage, I believe to mimic what delegates were doing. We don't need to do this anymore!

NathanReb:
We need to fix Cmdliner's warning first!

CHANGES.md
```diff
@@ -10,6 +10,8 @@
 
 ### Removed
 
+- Remove support for delegates (#428, @NathanReb)
```
pitag-ha, line 13: Maybe we can mention `delegate-info` here just to make sure that folks who were using delegates know how to adapt to the new situation? Something like "(...) You can use the `delegate-info` workflow instead."

NathanReb: Very good point, I'll add that! I'm also wondering if we shouldn't have named this command differently as delegate might not be meaningful for most users but that's a separate matter! We can still add another alias and remove `delegate-info` in its favor later on!

NathanReb:
I'll rebase on #429 so we can see the CI!

Leonidas-from-XIV:
#429 has been merged and we released 1.6.1 with the fix since some people needed these fixes.

bin/help.ml
```diff
@@ -78,17 +78,15 @@ let release =
...
-        \        by other systems hence do not try to alter the archive and\n\
-        \        republish it with a different bit-stream after that point (if\n\
-        \        you are tempted to do this please consider taking a functional\n\
-        \        programming course). At worst 410 the archive from\n\
-        \        the WWW. But in most cases, if there is a problem with the\n\
-        \        archive, simply leave it there and publish a new one with an\n\
-        \        increased patch version number.";
+        "The distribution is now public. It may already have been picked up by \
+         other systems hence do not try to alter the archive and republish it \
+         with a different bit-stream after that point. At worst 410 the \
```
Leonidas-from-XIV, line 86: Since this is being touched: I would replace "bit-stream" with "content" because that's a more common term and is more understandable.

bin/help.ml
```diff
@@ -78,17 +78,15 @@ let release =
...
+         archive from the WWW. But in most cases, if there is a problem with \
```
Leonidas-from-XIV, line 87: I would also suggest changing the note about 410'ing from the WWW with something along the lines of:

> In case you want to change it, it is suggested to make a new release with a new version instead

Since Github doesn't give you access to HTTP codes anyway.

Leonidas-from-XIV (approved):
Excellent! This will also make our code more reliable since we don't have poorly tested delegate support.

Looking forward to this PR kicking off dune-release 2.x development!

bin/help.ml
NathanReb: Good point, I'll change that!

PR #429, opened 2022-02-11
Fix compat with cmdliner 1.1+

NathanReb:


CHANGES.md
```diff
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Fix compatibility with recent cmdliner
```
Leonidas-from-XIV, line 11:
```suggestion
- Fix compatibility with Cmdliner 1.1.0 (#429, @NathanReb)
```

NathanReb:
> We should however also bump the minimum OCaml version to 4.08 since that's what Cmdliner requires.

You 're right that we should probably do it but I don't think that's the reason why. If we do, we should do it in a separate PR at the very least and make it an occasion to use new syntax such as letop!

Leonidas-from-XIV:
I get your point but at the moment the CI just doesn't run on 4.06 and 4.07 so if we merge it we basically have no guarantees that the code we write actually works on anything before 4.08 (say, if a future cmdliner 1.1.1 decides to support 4.06 again), which is not a good place to be in.

Nevertheless I think it would be nice to convert to use the let operators and e.g. remove any stdlib polyfills we might have, but I suggest that as a separate PR.

gpetiot:
Could we cut a minor release with this compat? This is needed to release projects that already upgraded cmdliner (like `ocamlformat`)

pitag-ha (approved):
Thanks, looks all good to me!

It's a bit of a pity to drop support for some compilers because of this, but to be fair there are several workarounds that still allow folks to use dune-release even if they really want to use an old compiler: they could have their project on a local switch and dune-release on the default global switch with compatible compiler version, or they could compile dune-release on a different switch and move the binary, or similar.

PR #431, opened 2022-03-08
RFC: Rely on the remote URL to get the main package name

gpetiot:
Based on the discussion in #366

The idea is to not check the dune-project file and the opam files as it is currently incorrect, but rely on git remote URL instead. Asking for some feedback before going further with this PR. (yeah it somehow overwrites the local config, will look into it if you like the idea)

PS: I'm doing this because releasing ocamlformat is painful because of this issue.

### DIFF
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -267,20 +267,6 @@ let infer_pkg_names dir = function
       else Ok (List.map (fun p -> Fpath.(basename @@ rem_ext p)) opam_files)
   | x -> Ok x
 
-let infer_from_opam_files dir =
-  infer_pkg_names dir [] >>= fun package_names ->
-  let shortest =
-    match package_names with
-    | [] -> assert false
-    | first :: rest ->
-        List.fold_left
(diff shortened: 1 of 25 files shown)


Leonidas-from-XIV:
We discussed this and agree that #366 is a problem. Our solution though would be to infer it from `dev-repo`, which is then the only thing that dune-release cares about. Having this in `remote.origin` is fragile and relies on the fact that people call it `origin` (or git does b default), which can be invalid (I frequently have multiple remotes with different names, because I check out other peoples remotes or switch which my remote is etc).

So if you want, you could change to use `dev-repo`, otherwise I'll make a PR for it.

gpetiot:
I tried to do this at first but I encoutered some dificulties because we couldn't read the opam file fields before building the `Pkg.t`, but building it required to compute the name first. But maybe I missed something indeed.

Leonidas-from-XIV:
Yes, it is possible that some bits need to be rewritten extensively, but that seems to be the least fragile (and especially "magic") solution for now :-/

PR #445, opened 2022-05-25
Fix dune-project name parsing

emillon:
The current parsing code looks for `(name X)` on all lines of `dune-project`, and returns the last one that matches. This can return the wrong value when opam file generation is used, for example:

    (lang dune 2.4)
    (name main)
    (generate_opam_files true)

    (package
     (name main))

    (package
     (name other))

In this case, the project name is parsed as `other`.

This PR fixes it by only looking at the beginning of the line and returning the first match. It also uses `Re` (already used) instead of direct string manipulation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@
 
 - Fix project name detection from `dune-project`. The parser could get confused
   when opam file generation is used. Now it only considers the first `(name X)`
-  in the file. (#<PR_NUMBER>, @emillon)
+  in the file. (#445, @emillon)
 
 ### Removed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 
 ### Fixed
 
+- Fix project name detection from `dune-project`. The parser could get confused
+  when opam file generation is used. Now it only considers the first `(name X)`
+  in the file. (#<PR_NUMBER>, @emillon)
+
 ### Removed
 
 - Remove support for delegates.
--- a/lib/pkg.mli
+++ b/lib/pkg.mli
@@ -155,18 +155,18 @@ val version_of_changelog : t -> Version.Changelog.t -> Version.t
 val dev_repo : t -> (string option, Sos.error) result
 (** dev-repo field with the ["git+"] prefix removed. *)
 
+(** {1 Dune project} *)
+
+val dune_project_name : Fpath.t -> (string option, [> Rresult.R.msg ]) result
+(** Returns the name stanza entry of the dune project, if any; else, returns
+    `None`. Returns an error, if one of the system calls fails. *)
+
 (**/**)
 
 val version_line_re : Re.t
 
 val prepare_opam_for_distrib :
   version:Version.t -> content:string list -> string list
-
-(** {1 Dune project} *)
-
-val dune_project_name : Fpath.t -> (string option, [> Rresult.R.msg ]) result
-(** Returns the name stanza entry of the dune project, if any; else, returns
-    `None`. Returns an error, if one of the system calls fails. *)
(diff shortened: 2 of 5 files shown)


emillon:
I went with `Re` because it's already available, but I can also revert to a more conservative fix if needed.

tests/bin/distrib-name-dune-project/run.t
```diff
@@ -0,0 +1,32 @@
+When a package name is inferred from dune-project, it should only use the
```
NathanReb, line 1: Ideally I would have preferred a unit test for this but we'll settle for a cram test.

CHANGES.md
```diff
@@ -8,6 +8,10 @@
 
 ### Fixed
 
+- Fix project name detection from `dune-project`. The parser could get confused
+  when opam file generation is used. Now it only considers `(name X)` at the
+  beginning of a line. (#445, @emillon)
```
Leonidas-from-XIV, line 13: I was a bit confused by this because I expected to see a `^`-equivalent in the regex, but I think what it means to say is "on its own line"?

tests/bin/distrib-name-dune-project/run.t
```diff
@@ -0,0 +1,32 @@
...
+  $ dune-release tag --yes
+  [-] Extracting tag from first entry in CHANGES.md
+  [-] Using tag "1.0.0"
+  [+] Tagged HEAD with version 1.0.0
+
+  $ dune-release distrib --dry-run --skip-lint --skip-build --quiet
+  $ ls _build
+  first-1.0.0
+  first-1.0.0.build
+  first-1.0.0.tbz
```
Leonidas-from-XIV, line 32: Could you add a test for

```
cat > dune-project << EOF
  > (lang dune 2.7)
  >  (name first)
```
To see what the semantics are if people hand-write these files and accidentally add whitespace?

Leonidas-from-XIV (commented):
This looks like a useful improvement and I guess for now we can just merge it (after clearing up the semantics). If we decide to change the behaviour (as we planned) we can revisit the code, let's not make perfect the enemy of good :+1:

NathanReb (approved):
Thanks for fixing this!

We'll need to backport it and release a 1.x version with this.

It's likely that we will drop all the dune-project parsing for 2.0.

CHANGES.md
emillon: Ah right I forgot to add `bol`. Well, "first match" might be a better heuristic anyway. Should I document it as such?

tests/bin/distrib-name-dune-project/run.t
emillon: I can do that

CHANGES.md
emillon: done!

tests/bin/distrib-name-dune-project/run.t
emillon: done!

tests/bin/distrib-name-dune-project/run.t
emillon: done!

Leonidas-from-XIV (approved):

lib/pkg.mli
```diff
@@ -168,6 +168,8 @@ val dune_project_name : Fpath.t -> (string option, [> Rresult.R.msg ]) result
 (** Returns the name stanza entry of the dune project, if any; else, returns
     `None`. Returns an error, if one of the system calls fails. *)
 
+val dune_project_name_string : string -> string option
```
NathanReb, line 171: Probably belongs in the undocumented section below!

PR #458, opened 2023-03-09
Add flag to `dune-release check` that attempts to discover and parse the change log

gridbugs:
I found that a project where `dune-release check` succeeded didn't publish because its change log failed to parse, and iterating on debugging and fixing the change log was difficult with no simple way to exercise the change log parsing logic.

### DIFF
--- a/tests/bin/check/run.t
+++ b/tests/bin/check/run.t
@@ -223,3 +223,47 @@ The [--working-tree] option used so far, makes `check` be run on the working tre
   Have you provided a github uri in the dev-repo field of your main opam file? If you don't use github, you can still use dune-release for everything but for publishing your release on the web. In that case, have a look at `dune-release delegate-info`.
   [FAIL] The dune project doesn't contain a name stanza. Please, add one.
   [2]
+
+Detect case where there is no changelog file present:
+  $ dune-release check --skip-lint --skip-build --skip-tests --working-tree --check-change-log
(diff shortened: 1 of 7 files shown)


Leonidas-from-XIV:
Definitely an useful addition, but for a more consistent user experience I think it would make sense to flip it: add `--skip-change-log-check` to match the other `--skip-*` flags and make this an opt-out behavior.

gridbugs:
Good suggestion. Fixed!

tests/bin/check/run.t
```diff
@@ -41,13 +51,16 @@ If the condition described above is fulfilled, there are 4 checks to be performe
   [-] Performing lint for package my_pkg in <test_directory>
   [FAIL] File README is missing.
   [FAIL] File LICENSE is missing.
-  [FAIL] File CHANGES is missing.
+  [ OK ] File CHANGES is present.
```
Leonidas-from-XIV, line 54: Please keep one test where the changelog is missing, so this test case doesn't fall by the wayside.

Leonidas-from-XIV (changes requested):
Very nice. Just a small note that we should still keep a case where the changelog file is missing, besides that it's ready to go!

tests/bin/check/run.t
gridbugs: Fixed

PR #459, opened 2023-03-09
Add subcommand to print change log entry

gridbugs:
I did a release with dune-release but found that the change log in the PR was empty to my surprise. I want an easy offline way to check the tag and change log entries pulled from the change log file.

### DIFF
--- a/bin/change_log.ml
+++ b/bin/change_log.ml
@@ -4,55 +4,31 @@
    %%NAME%% %%VERSION%%
   ---------------------------------------------------------------------------*)
 
+open Bos_setup
+open Dune_release
+
+let print_latest_change_log_entry () =
+  let pkg = Pkg.v ~dry_run:true () in
+  Pkg.change_log pkg >>= Text.change_log_file_last_entry
+  >>| (fun (version, (title, body)) ->
+        Fmt.pr "Tag: %a\nTitle: %s\nBody:\n%s\n" Version.Changelog.pp version
+          title body;
+        0)
+  |> Cli.handle_error
+
 open Cmdliner
 
-let cmds =
-  [
-    Tag.cmd;
-    Distrib.cmd;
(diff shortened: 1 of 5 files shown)


bin/change_log.ml
```diff
@@ -0,0 +1,47 @@
...
+   %%NAME%% %%VERSION%%
+  ---------------------------------------------------------------------------*)
+
+open Bos_setup
+open Dune_release
+
+let print_latest_change_log_entry () =
+  let pkg = Pkg.v ~dry_run:true () in
+  Pkg.change_log pkg >>= Text.change_log_file_last_entry
+  >>| (fun (version, (title, body)) ->
```
Leonidas-from-XIV, line 13: I was about to point out that it would be nice to use the `let*` operators, but given all the other entrypoints are also using infix operators it's ok too. Up to you.

tests/bin/change-log/dune
```diff
@@ -0,0 +1,8 @@
+(env
+ (_
+  (binaries
+   (../helpers/make_dune_release_deterministic.exe as
+     make_dune_release_deterministic))))
```
Leonidas-from-XIV, line 5: Is this binary required? I don't see it being used in the cram test.

Leonidas-from-XIV (commented):
The idea is good, but sure an additional command is the way to go. Given you were surprised by the empty changelog, maybe just printing the parsed changelog during `dune-release tag` would be a better UX? In this case the user has to run the subcommand specifically to check and if they don't, then they won't notice that the change log is empty. 

In a way this feels more like a debug command that we parse the changelog correctly than a command that would be helpful for users.

gridbugs:
I agree with you point that this would be more useful as part of `dune-project tag`. I'll close this PR and open a new PR that adds that behaviour.

PR #460, opened 2023-03-09
Treat '-' underlines as H1 for first heading

gridbugs:
In changelogs, people can accidentally use '-' instead of '=' for markdown titles. Currently if they do this the file fails to parse if they also use '##' headings for versions, because the parser assumes that version headings will have higher heading values than the file's title.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- In markdown changelog files, treat underline made of '-' characters as a H1 heading
+  rather than H2 heading if it is the first heading in a file (#460, @gridbugs)
+
 ### Deprecated
 
 ### Fixed
--- a/lib/text.ml
+++ b/lib/text.ml
@@ -34,7 +34,7 @@ let underline_header n uchar l before rest =
     | None -> None
     | Some t -> Some (n, strf "%s\n%s" t l, List.tl before, rest)
 
-let rec find_markdown_header before = function
+let rec find_markdown_header ~first before = function
   | [] -> None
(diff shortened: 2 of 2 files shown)


gridbugs:
This is possibly a controversial change so I want to see what people think before spending time writing tests.

Leonidas-from-XIV (commented):
The code looks fine to me, but can you point at changelogs that have this issue?

I am wondering whether being liberal about the headers could lead to unexpected lines being picked up as version numbers.

gridbugs:
> The code looks fine to me, but can you point at changelogs that have this issue?

I noticed this problem when adding https://github.com/ocaml-batteries-team/batteries-included to opam-overlays but I sent them a PR to fix their changelog.

> I am wondering whether being liberal about the headers could lead to unexpected lines being picked up as version numbers.

Yeh now that I've had more time to think about this I don't think this is the right solution. I've since run into a changelog in the wild that uses '-' underlines for its version numbers and has no heading which I think would be incorrectly parsed if we merged this.

I think what we need instead is a specific error message when parsing fails for the reason I described in the commit message.

Leonidas-from-XIV:
> I think what we need instead is a specific error message when parsing fails for the reason I described in the commit message.

Yes, I think I would agree. We added the [keepachangelog](https://keepachangelog.com/en/1.1.0/) format support but I don't think we should aim to support every type of changelog. Or if we do, it should probably be spun off as a separate library with the explicit goal of supporting as many changelog formats as feasible, similar to [vandamme](https://github.com/tech-angels/vandamme/).

PR #465, opened 2023-04-18
Use project name for PR title and branch

emillon:
If we're submitting a PR for several packages, this uses "project name (N packages)" as PR title, and the project name is used for the branch.
Previously the branch was taken from the shortest package name.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,9 +4,11 @@
 
 ### Changed
 
-- Running `dune-release check` now attempts to discover and parse the
-  change log, and a new flag `--skip-change-log` disables this behaviour.
-  (#458, @gridbugs)
+- Running `dune-release check` now attempts to discover and parse the change
+  log, and a new flag `--skip-change-log` disables this behaviour. (#458,
+  @gridbugs)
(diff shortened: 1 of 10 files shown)


emillon:
NB the description in the PR is still wrong, it says "XDG Base Directory Specification" for dune.

Leonidas-from-XIV:
I've added some code to handle `-p` (it's not very nice but the whole code packaging code is an utter mess).

The issue why it picks the description of XDG is because it uses `Pkg.opam_descr` is because it uses `List.hd pkgs` and that is of course utterly random and wrong. I wonder if we should revive #447 to clean up, because attempting to fix this here is just patching leaks.

emillon:
Ah, I didn't see your comment about #447 so I implemented the "find the main package" in a different way. Let me know what you think (I used this for https://github.com/ocaml/opam-repository/pull/23748)

Leonidas-from-XIV:
I guess this works as well. It is better than the status quo and if we realize this has some particular issues it can be fixed up later on. It does need a changelog entry though.

Closes #477. (Wonder if this works from a comment?)

PR #466, opened 2023-04-18
Implement support for OPAM files in `opam/` folder

Leonidas-from-XIV:
This requires the as of now not yet released dune 3.8, hence a draft. CI will most likely fail the test.

### DIFF
--- a/dune-release.opam
+++ b/dune-release.opam
@@ -18,8 +18,9 @@ license: "ISC"
 homepage: "https://github.com/tarides/dune-release"
 bug-reports: "https://github.com/tarides/dune-release/issues"
 depends: [
-  "dune" {>= "2.7"}
   "ocaml" {>= "4.08.0"}
+  "dune" {>= "2.7"}
+  "dune" {>= "3.8" & with-test}
   "curly" {>= "0.3.0"}
   "fmt" {>= "0.8.7"}
   "fpath" {>= "0.7.3"}
--- a/lib/pkg.ml
+++ b/lib/pkg.ml
@@ -91,7 +91,13 @@ let readme p =
 let opam p =
   match p.opam with
   | Some f -> Ok f
-  | None -> name p >>| fun name -> Fpath.v (name ^ ".opam")
+  | None ->
+      name p >>= fun name ->
+      let filename = name |> Fpath.v |> Fpath.add_ext ".opam" in
+      let opam_subdir = Fpath.v "opam" in
+      let in_opam_subdir = Fpath.append opam_subdir filename in
+      OS.File.exists in_opam_subdir
+      |> Result.map (function true -> in_opam_subdir | false -> filename)
 
 let opam_descr p =
   let descr_file_for_opam opam =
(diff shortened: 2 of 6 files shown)


Alizter:
Any news on this?

Leonidas-from-XIV:
@Alizter Ah right, now with dune 3.8 out of the door, let me bump the dune dependency.

dune-release.opam
```diff
@@ -18,7 +18,7 @@ license: "ISC"
 homepage: "https://github.com/tarides/dune-release"
 bug-reports: "https://github.com/tarides/dune-release/issues"
 depends: [
-  "dune" {>= "2.7"}
+  "dune" {>= "3.8"}
```
Alizter, line 21: Will this mean people will have to use an older version of dune-release for older dune?

Leonidas-from-XIV: No, dune-release is essentially just a stand-alone binary (that just happens to often be installed in the switch thus messing with users dependencies) which only uses very basic Dune commands (pretty sure it would work on projects using Dune 1.0, though we don't have tests for that). The support for the opam subfolder doesn't require any specific support from Dune; the only reason I've bumped it is to get Dune 3.8 in the tests so the feature itself can be tested.

In theory the tests could be rewritten to not use the Dune 3.8 support and just create the opam files manually, but given updating Dune has little downside I think this approach is fair enough and can also serve as an example how to apply this to one's own projects.

Leonidas-from-XIV:
The CI failure on 5.1 is due to odoc and I submitted a PR: https://github.com/ocaml/opam-repository/pull/23924

dune-release.opam
emillon: Yeah, I think that's fine.

lib/pkg.ml
```diff
@@ -89,7 +89,15 @@ let readme p =
 let opam p =
   match p.opam with
   | Some f -> Ok f
-  | None -> name p >>| fun name -> Fpath.v (name ^ ".opam")
+  | None -> (
+      name p >>= fun name ->
+      let filename = name |> Fpath.v |> Fpath.add_ext ".opam" in
+      let opam_subdir = Fpath.v "opam" in
+      let in_opam_subdir = Fpath.append opam_subdir filename in
+      match OS.File.exists in_opam_subdir with
```
emillon, line 97: There's probably a `Rresult.R.map` or `Result.map` to use if you'd like to save a line

lib/pkg.ml
```diff
@@ -379,11 +399,24 @@ let prepare_opam_for_distrib ~version ~content =
   let without_version = List.filter is_not_version_field content in
   Fmt.str "version: \"%a\"" Version.pp version :: without_version
 
+let opam_file_contents name =
+  let filename = name |> Fpath.v |> Fpath.add_ext ".opam" in
+  let location =
+    OS.File.exists filename >>= function
```
emillon, line 405: given you're already 4.08+ you can take the opportunity to use `let*`/`let+` instead of `>>=`/`>>|` for new code, but that's not mandatory either

dune-release.opam
emillon: Also, I think that you just require a dune 3.8 binary for tests. So you can depend on 2.7, keep lang 2.7, but test-depend on dune >= 3.8 with-tests.

tests/bin/opam-file-locations/run.t
```diff
@@ -0,0 +1,100 @@
...
+  [-] Running package tests in _build/myproject-0.1.0
+  [ OK ] package(s) pass the tests
+  
+  [+] Distribution for myproject 0.1.0
+  [+] Commit <deterministic>
+  [+] Archive _build/myproject-0.1.0.tbz
+
+Now let's move the `.opam` file to the `opam/` subfolder. OPAM supports `.opam`
+files in the `opam/` subfolder, but for dune to pick it up we need to tell it
+to look in that folder.
```
emillon, line 52: That's very clear. Does dune 3.8 support generating files in `opam/` ? That could also be a nice e2e test.

dune-release.opam
Leonidas-from-XIV: That's a cool idea, although specifying this in `dune-project` is a bit tricky as it injects the `dune` dependency in a wrong way:

```
(dune (and (>= 3.8) :with-test)
```

yields

```
`"dune" {>= "2.7" & >= "3.8" & with-test}
```

which I think is totally wrong since it would then have dune as a test-only dependency. OTOH

```
(dune (or (>= 2.7) (and (>= 3.8) :with-test))
```

yields

```
"dune" {>= "2.7" & >= "2.7" | >= "3.8" & with-test}
```

which *might* be correct but depends on how strong `|` binds. The [manual](https://opam.ocaml.org/doc/Manual.html#Package-Formulas) is not very explicit about it.

emillon: You're hitting #3431 unfortunately. As a workaround, I think that you can add an explicit bound on dune, and the test one, which should generate 2 clauses and have the correct semantics. But might not be worth it, so we can put that aside.

tests/bin/opam-file-locations/run.t
Leonidas-from-XIV: It does (and it works)! I was a bit on the fence, because it feels a bit like testing dune and not dune release, but I guess this won't hurt :)

This made me think of a future where the repos would not have the `opam` files pregenerated but rather dune-release would create them just for the tarball. Though that introduces some issues with pinning.

dune-release.opam
Leonidas-from-XIV: Ah, I knew I have seen this issue mentioned before :) Your workaround seems to work perfectly and dune will keep the dune dependency at the right minimal version in the first bound while leaving the second one alone! Neat.

emillon: yes, I've used that before. one thing to consider is that when you'll remove this, you need to remember to remove the dune dep otherwise it will not generate the constraint that corresponds to dune lang

PR #469, opened 2023-05-03
opam: fix lower bounds

emillon:
We use `Rresult.R.failwith_error_msg` which has been added in 0.6.0.


Leonidas-from-XIV (approved):
Looks good, thanks for the fix!

emillon:
This does not fix the lower bounds run, I was debugging the cram failure. I'll send a followup PR

PR #470, opened 2023-05-04
Fix lower bound against bos

emillon:
We require the fixes in 0.1.1 for command running and the printing fixes in 0.1.3.


emillon:
@Leonidas-from-XIV here we are. This finishes #469. I thought the last bit was a dune 2 vs 3 issue but it's actually caused by `bos`.

Leonidas-from-XIV (approved):

PR #478, opened 2023-06-27
Upgrade to ocamlformat 0.26.0

Julow:
Mostly due to the change in indentation of a `let` or `if` inside parentheses.

The aim of this commit is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Upgrade to OCamlformat 0.26.0
+7f4890bc747650982e5f01eb1e570c9fa127bbdb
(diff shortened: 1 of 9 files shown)


Leonidas-from-XIV:
Thanks @Julow!

My personal, very subjective opinion is that I like the change since I find `if-then-else` overall rather messy and hard to read so this is a welcome improvement. All the changed blocks read better than before as it is more apparent what is part of which branch of the `if`.

Can you ping me when a final version is released? I'd welcome upgrading to the new version and reformatting the source code.

Julow:
Thanks for the feedback! I'll update this PR when the release is ready.

Julow:
OCamlformat 0.26.0 has been released! This PR can be merged (not squashed due to `.git-blame-ignore-revs`) or closed if it's a bit too early to upgrade.

Issue #479, opened 2023-07-18
Don't try to understand my Github URL


Julow:
I get this error when I run `dune-release opam submit`:

```
dune-release: [ERROR] The URL to your remote fork of opam-repository gh:Julow/opam-repository does not seem to point to a github repo.
  Try editing your config with `dune-release config set remote <URL>` or providing a valid Github repo URL via the --remote-repo option.
```

No other URL exists due to my configuration. SSH URLs can be arbitrarily complex (even when pointing to Github) and I don't want any tool to decode mines. Git doesn't, for example.

Leonidas-from-XIV:
Does the workaround mentioned work though?

The whole `opam submit` thing is a bit wonky since it uses a local checkout, edits it and submits. In the future I would much prefer to create a commit using the Github API. Then there is no need to deal with peoples URLs and setups.

Julow:
The problematic part is not requiring a URL, which works fine as long as it doesn't have a meaning other than being a URL for git.
API calls require a user name, and it's currently decoded from the URL of the fork, which makes sense in many cases but doesn't work in general.

I have not found a workaround, this requires a fix: https://github.com/tarides/dune-release/pull/480

Julow:
The user name is used to identify branches when creating the PR, something like `{ "head": "user:branch-name" }`.

Julow:
This is partly fixed by https://github.com/tarides/dune-release/pull/480 but it's not documented and the tool will not ask for the information interactively (instead it'll fail for a bad reason).

Julow:
The workaround added in https://github.com/tarides/dune-release/pull/480 doesn't work anymore:
```
[!] The user configuration field is deprecated and will be removed in 2.0.0 as it is redundant with the remote field. Setting it to the wrong value can lead to bugs. Please use the remote field only.
```
How can I stop dune-release from parsing a string that cannot be parsed ?

Julow:
Sorry for the noise. I was using the released version by mistake.
The work around continues to work:
```
dune-release opam submit --user my_user -r my_url -p ...
```

PR #480, opened 2023-07-18
Opam: Use the 'user' option as the fork owner

Julow:
This is an attempt at fixing https://github.com/tarides/dune-release/issues/479 but I noticed that the `user` option is deprecated.

Decoding the remote URL can easily fail, instead let the user provide the correct username to use in API calls.

Only attempt to decode the remote URL if the `user` option is not set. The error message is changed to encourage setting the `user` option instead of asking users to change their URLs.

### DIFF
--- a/bin/opam.ml
+++ b/bin/opam.ml
@@ -164,19 +164,28 @@ let open_pr ~dry_run ~changes ~remote_repo ~fork_owner ~branch ~token ~title
         | Ok () -> Ok 0
         | Error _ -> msg ())
 
-let parse_remote_repo remote_repo =
-  match Github_repo.from_uri remote_repo with
-  | Some repo -> Ok repo
-  | None ->
-      R.error_msgf
-        "The URL to your remote fork of opam-repository %s does not seem to \
-         point to a github repo.\n\
-         Try editing your config with `dune-release config set remote <URL>` \
-         or providing a valid Github repo URL via the --remote-repo option."
-        remote_repo
+(** Get the value from the [user] config, if it is unset try to parse it from
+    the [remote_repo]. *)
+let get_fork_owner ~fork_owner ~remote_repo =
+  match fork_owner with
+  | Some o -> Ok o
+  | None -> (
+      match Github_repo.from_uri remote_repo with
+      | Some repo -> Ok repo.owner
+      | None ->
+          R.error_msgf
+            "The owner of the opam-repository fork on Github is needed but \
+             couldn't be parsed from the remote fork.\n\
+             Please configure your Github user name with `dune-release config \
+             set user <owner>` or providing it via the\n\
+            \             `--user` option.")
 
-let submit ~token ~dry_run ~yes ~opam_repo ~pkgs_to_submit local_repo
-    remote_repo pkgs auto_open ~draft =
+let submit ~token ~dry_run ~yes ~opam_repo ~pkgs_to_submit
+    {
+      Config.Opam_repo_fork.local = local_repo;
+      remote = remote_repo;
+      user = fork_owner;
+    } pkgs auto_open ~draft =
   List.fold_left
     (fun acc pkg ->
       get_pkg_dir pkg >>= fun pkg_dir ->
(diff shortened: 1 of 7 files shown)


Leonidas-from-XIV:
Hmm, yeah, this approach in general seems reasonable to me. I think the idea was that for 80% of users no configuration is required and it can be inferred but you're right that it is not a 100% solution so being able to explicitly provide the user sounds like a good option to me.

Leonidas-from-XIV:
> This might be fixed by your proposal of using the API instead of a local repo in https://github.com/tarides/dune-release/issues/479 ?

I do hope so, however depending on whether we still want to support local repos (I would say no, others might disagree) this might not be a complete solution.

Julow:
This has slowed me down in release Odoc today.

gpetiot:
I dug around and I didn't find anything in the API that would allow to do that without the username/ownername (with just the url as you suggested @Leonidas-from-XIV), what we are doing right now (https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request) looks like the only way.

@Leonidas-from-XIV do you have more details about what you suggested, or more concerns about this PR?
From where I'm at, it looks like the best solution to #479, and we could merge it.

bin/opam.ml
```diff
@@ -164,19 +164,28 @@ let open_pr ~dry_run ~changes ~remote_repo ~fork_owner ~branch ~token ~title
...
-         point to a github repo.\n\
-         Try editing your config with `dune-release config set remote <URL>` \
-         or providing a valid Github repo URL via the --remote-repo option."
-        remote_repo
-
-let submit ~token ~dry_run ~yes ~opam_repo ~pkgs_to_submit local_repo
-    remote_repo pkgs auto_open ~draft =
+(** Get the value from the [user] config, if it is unset try to parse it from
+    the [remote_repo]. *)
+let get_fork_owner ~fork_owner ~remote_repo =
```
Leonidas-from-XIV, line 169: Given that this function does only return `fork_owner` when it is set, I think the more sensible thing is to not call it when `fork_owner` is set.

bin/opam.ml
```diff
@@ -215,7 +224,7 @@ let submit ~token ~dry_run ~yes ~opam_repo ~pkgs_to_submit local_repo
     | Some { owner; repo } -> Text.rewrite_github_refs ~user:owner ~repo changes
     | None -> changes
   in
-  parse_remote_repo remote_repo >>= fun { owner = fork_owner; _ } ->
+  get_fork_owner ~fork_owner ~remote_repo >>= fun fork_owner ->
```
Leonidas-from-XIV, line 227: Probably better to just pass in `fork_owner` here.

Leonidas-from-XIV (changes requested):
Can you also add a changelog entry to the PR?

bin/opam.ml
Julow: Checking on `fork_owner` is the only intent behind this function, to avoid making the larger function even larger. Do you suggest inlining it all ?

bin/opam.ml
Julow: Before this point, `fork_owner` is an optional value that is probably `None` for most users. After this point, `fork_owner` is a string.
Inlining part of `get_fork_owner` here wouldn't be very nice to read. How do you suggest writing that ?

bin/opam.ml
Leonidas-from-XIV: Yes, because if you pass `fork_owner` you already know the result. No need to call `get_fork_owner` to tell you that the fork owner you had already is indeed the fork owner.

This would be different if there was something to be calculated from the owner so passing in an optional pre-calculated owner was important.

bin/opam.ml
Leonidas-from-XIV: An option could make a function along the lines `val fork_owner_if_unknown : fork_owner:string option -> _` that calls `parse_remote_repo` if `fork_owner` is unknown.

gpetiot:
I modified the code according to the review and added an entry in the changelog.

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
LGTM, thanks!

bin/opam.ml
Julow: This was exactly the intent behind `get_fork_owner`.

bin/opam.ml
```diff
@@ -164,19 +165,25 @@ let open_pr ~dry_run ~changes ~remote_repo ~fork_owner ~branch ~token ~title
         | Ok () -> Ok 0
         | Error _ -> msg ())
 
-let parse_remote_repo remote_repo =
+(** Get the value from the [user] config, if it is unset try to parse it from
```
Julow, line 168: The docstring is now wrong.

Julow (commented):
Thanks!

PR #486, opened 2024-02-12
Update tests for dune 3.14

emillon:
dune.3.14.0 adds a loc in its output, see ocaml/dune#9851.

This updates the test output and the constraint.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -28,8 +28,8 @@ with [Dune](https://github.com/ocaml/dune) and hosted on
   ;; two dependencies on dune to work around
   ;; https://github.com/ocaml/dune/issues/3431
   dune
-  ;; the tests require dune 3.8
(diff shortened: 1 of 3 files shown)


emillon:
I tried to `grep -v` to be compatible with both old and new dune, but this is tricky because the line is on stderr, but there are (not pretty) alternatives.

Leonidas-from-XIV (approved):
Looks ok, given 3.8 and 3.14 support the same OCaml this is close enough.

I assume we'll need add the dune 3.14 constraint in `:with-test` dependencies on opam-repository as well?

emillon:
I already updated the constraints for 2.0.0 here: https://github.com/ocaml/opam-repository/pull/25238

Leonidas-from-XIV:
Merging as dune 3.14 has been released.

PR #497, opened 2025-01-29
drop migration path for the config file

v-gb:
Currently, it systematically fails:

```shell
$ dune-release 
[-] Upgrading configuration files: /home/foo/.dune => /home/foo/.config/dune
dune-release: [ERROR] move /home/foo/.dune to /home/foo/.config/dune/release.yml: Destination exists
```

when using dune package management, due to dune getting installed to
~/.dune/bin/dune.

Considering that:

- the migration code from 2018
- it doesn't work at all, since aside from failing unnecessarily, it would
  also mess things up by renaming the directory ~/.dune to
  ~/.config/dune/dune/release.yml

I see no reason to salvage this code.

The second commit cleans up some Result.t that have become infallible.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
 
 ### Fixed
 
+- Make `dune-release` not fail in the presence of `~/.dune/bin/dune` (which is present
+  when using dune package management)
+
 ### Removed
 
 ### Security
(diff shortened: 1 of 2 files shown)


v-gb:
Right, but it doesn't remove any previous installed one, and no reason to keep nonfunctional code.

samoht:
Thanks!

PR #499, opened 2025-02-03
bistro: stop publishing docs

v-gb:
There is little reason to publish docs to github, considering ocaml.org already makes api docs available. Changing this behavior allows setting the opam doc field to any url without dune-release choking on it.

Maybe it'd be preferable to delete the docs publishing code entirely?


samoht:
Thanks! And yes, deleting the docs publishing code enterily seems like a good idea

samoht:
Closing in favour of #500 - is there more dead-code to remove?

PR #502, opened 2025-02-03
Prepare the 2.1 release

samoht:


Leonidas-from-XIV:
Should that not be a 3.0 release, since it changes pretty major functionality that some people might still rely on?

PR #506, opened 2025-04-25
Add a dune package management GHA workflow

Sudha247:
This introduces a workflow for dune package management. It would help us ensure the package builds with dune developer preview


Sudha247:
The idea here was to do the build and tests with `dune pkg`, without using `opam`. But it seems the tests themselves require `opam` for linting, etc. I'm closing this PR for now, we can revisit this later when looking at release management with `dune pkg`.

PR #509, opened 2025-08-29
Use opam libs instead of `opam` CLI to lint

Leonidas-from-XIV:
As far as I am aware the main thing `dune-release` requires `opam` (the CLI binary) for is linting. Given `dune-release` already depends on the opam libs and the opam libs helpfully provide an API for linting, I thought I'd take a shot at implementing the linting using the library instead of the CLI. This has the additional advantage that to use `dune-release` normally there is no need to install `opam` anymore.

It seems the change is fairly trivial and actually removes a good amount of subprocess handling code. That said, it isn't well tested and it could be made nicer because the API outputs the issues in a nice way and we could log them in a nice way too.

The formatting might be a bit messed up because I backported the formatting from #508, I'll gladly rebase once that is merged.

cc @shonfeder

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -46,32 +46,6 @@ let lint_std_files ~dry_run pkg =
   in
   List.fold_left go 0 std_files
 
-let lint_file_with_cmd ~dry_run ~file_kind ~cmd ~handle_exit file errs =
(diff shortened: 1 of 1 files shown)


samoht:
Sounds like a good idea to not depend on the `opam` binary.

PR #510, opened 2025-09-01
Use newer GH runners for checks

Leonidas-from-XIV:
It seems that Github [discontinued the Ubuntu 20.04 runner](https://github.com/actions/runner-images/issues/11101) so this PR updates it to a less old one.


samoht:
Thanks!

PR #515, opened 2026-01-24
Support .gitattributes export-ignore in distribution archives

jberdine:
Parse .gitattributes files and exclude paths marked with the export-ignore
attribute from distribution tarballs. This is the mechanism used by
git-archive to exclude files, allowing projects to exclude dev-only files
like dune-workspace from releases.

Supported patterns:
- Exact matches: dune-workspace
- Directory patterns: .github/**
- Glob patterns: *.log, test_*, file?.txt
- Double star: **/build, src/**/test.ml

Paths are normalized before matching (handles ./ and ../).

Not supported:
- Escaped patterns (\\! for literal \!)
- Quoted patterns ("a b" for patterns with spaces)
- Case insensitivity (core.ignorecase)
- Negation patterns (\!pattern)
- Subdirectory .gitattributes files

Testing:
- Unit tests comparing behavior against git check-attr
- Tests cover pattern matching and parsing edge cases
- .gitattributes content is generated from test cases to ensure sync
- Archive integration test verifies end-to-end exclusion without requiring git

Signed-off-by: Josh Berdine <josh@berdine.net>

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,31 @@ The full documentation of this command is available with
 dune-release help distrib
 ```
 
+#### Excluding files with .gitattributes
+
+Files marked with the `export-ignore` attribute in `.gitattributes` will be excluded from the distribution archive. This can be used to exclude development files like `dune-workspace` that should not be included in releases.
+
+Example `.gitattributes`:
+```
+dune-workspace export-ignore
+.github/** export-ignore
+```
+
+**Supported patterns:**
+- Exact filenames: `dune-workspace`
+- Directory patterns: `.github/**` (matches all files under `.github/`)
+- Glob patterns: `*.log`, `test_*`, `file?.txt`
+- Double star in path: `**/build`, `src/**/test.ml`
+- Path normalization: handles `./` and `../` in paths
+
+**Not supported:**
+- Negation patterns (`!pattern`)
+- Escaped patterns (`\!` for literal `!`)
+- Quoted patterns (`"a b"` for patterns with spaces)
+- Character classes (`[abc]`)
+- Case insensitivity (`core.ignorecase`)
+- `.gitattributes` files in subdirectories
+
 
 ### Publish the distribution online
 
--- a/README.md
+++ b/README.md
@@ -127,13 +127,15 @@ dune-workspace export-ignore
 - Double star in path: `**/build`, `src/**/test.ml`
 - Path normalization: handles `./` and `../` in paths
 
-**Not supported:**
+**Unsupported pattern syntax** (such patterns are skipped with a warning):
 - Negation patterns (`!pattern`)
 - Escaped patterns (`\!` for literal `!`)
 - Quoted patterns (`"a b"` for patterns with spaces)
 - Character classes (`[abc]`)
-- Case insensitivity (`core.ignorecase`)
-- `.gitattributes` files in subdirectories
+
+**Other limitations:**
+- Only the repository-root `.gitattributes` is read; files in subdirectories are ignored.
+- Matching is always case-sensitive; `core.ignorecase` is not consulted.
 
 
 ### Publish the distribution online
--- a/lib/archive.ml
+++ b/lib/archive.ml
@@ -117,8 +117,15 @@ module Tar = struct
     String.concat (List.rev (end_of_file :: t))
 end
 
-let path_set_of_dir dir ~exclude_paths =
-  let not_excluded p = Ok (not (Fpath.Set.mem (Fpath.base p) exclude_paths)) in
+let path_set_of_dir dir ~exclude_paths ~export_ignore =
+  let not_excluded p =
+    if Fpath.Set.mem (Fpath.base p) exclude_paths then Ok false
+    else
+      match Fpath.rem_prefix dir p with
+      | None -> Ok true
+      | Some rel_path ->
+          Ok (not (List.exists (Gitattributes.matches rel_path) export_ignore))
+  in
   let traverse = `Sat not_excluded in
   let elements = `Sat not_excluded in
   let err _ e = e in
@@ -126,24 +133,29 @@ let path_set_of_dir dir ~exclude_paths =
   >>= OS.Path.fold ~dotfiles:true ~err ~elements ~traverse Fpath.Set.add
         Fpath.Set.empty
 
-let tar dir ~exclude_paths ~root ~mtime =
+let tar dir ~exclude_paths ~export_ignore ~root ~mtime =
   let tar_add file tar =
     let fname =
       match Fpath.rem_prefix dir file with
       | None -> assert false
       | Some file -> Fpath.(root // file)
     in
-    Logs.info (fun m -> m "Archiving %a" Fpath.pp fname);
     tar >>= fun tar ->
     OS.Dir.exists file >>= function
-    | true -> Tar.add tar fname ~mode:0o775 ~mtime `Dir
+    | true ->
+        (* Skip directories - they will be created implicitly when their
+           contents are added. This ensures that directories whose contents
+           are excluded via export-ignore patterns don't appear as empty
+           directories in the archive. *)
+        Ok tar
     | false ->
+        Logs.info (fun m -> m "Archiving %a" Fpath.pp fname);
         OS.Path.Mode.get file >>= fun mode ->
         OS.File.read file >>= fun contents ->
         let mode = if 0o100 land mode > 0 then 0o775 else 0o664 in
         Tar.add tar fname ~mode ~mtime (`File contents)
   in
-  path_set_of_dir dir ~exclude_paths >>= fun fset ->
+  path_set_of_dir dir ~exclude_paths ~export_ignore >>= fun fset ->
   Fpath.Set.fold tar_add fset (Ok Tar.empty) >>| fun tar -> Tar.to_string tar
 
 (* Bzip2 compression and unarchiving *)
--- a/lib/archive.ml
+++ b/lib/archive.ml
@@ -118,7 +118,7 @@ module Tar = struct
 end
 
 let path_set_of_dir dir ~exclude_paths ~export_ignore =
-  let not_excluded p =
+  let included p =
     if Fpath.Set.mem (Fpath.base p) exclude_paths then Ok false
     else
       match Fpath.rem_prefix dir p with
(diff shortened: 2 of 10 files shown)


lib/archive.ml
```diff
@@ -117,33 +117,45 @@ module Tar = struct
     String.concat (List.rev (end_of_file :: t))
 end
 
-let path_set_of_dir dir ~exclude_paths =
-  let not_excluded p = Ok (not (Fpath.Set.mem (Fpath.base p) exclude_paths)) in
+let path_set_of_dir dir ~exclude_paths ~export_ignore =
+  let not_excluded p =
```
Leonidas-from-XIV, line 121: To avoid the negation, maybe rename it to `included`. It's kind of hard to wrap one's head around the double negated name and then the negation in the code.

lib/archive.ml
```diff
@@ -117,33 +117,45 @@ module Tar = struct
...
     in
-    Logs.info (fun m -> m "Archiving %a" Fpath.pp fname);
     tar >>= fun tar ->
     OS.Dir.exists file >>= function
-    | true -> Tar.add tar fname ~mode:0o775 ~mtime `Dir
+    | true ->
+        (* Skip directories - they will be created implicitly when their
+           contents are added. This ensures that directories whose contents
+           are excluded via export-ignore patterns don't appear as empty
+           directories in the archive. *)
```
Leonidas-from-XIV, line 149: But isn't this what one would expect to do? If I exclude a directory, I would expect the directory to show up, but not its contents.

Git does this but its more to do with the fact that Git doesn't track directories at all, so an empty directory is unrepresentable. But `dune-release` doesn't have this issue.

lib/archive.mli
```diff
@@ -13,16 +13,19 @@ open Bos_setup
...
-(** [tar dir ~exclude_paths ~root ~mtime] is a (us)tar archive that contains the
-    file hierarchy [dir] except the relative hierarchies present in
-    [exclude_paths]. In the archive, members of [dir] are rerooted at [root] and
-    sorted according to {!Fpath.compare}. They have their modification time set
-    to [mtime] and their file permissions are [0o775] for directories and files
-    executable by the user and [0o664] for other files. No other file metadata
-    is preserved.
+(** [tar dir ~exclude_paths ~export_ignore ~root ~mtime] is a (us)tar archive
+    that contains the file hierarchy [dir] except:
+    - relative hierarchies present in [exclude_paths] (basename matching)
```
Leonidas-from-XIV, line 22: I wonder if it wouldn't make more sense to just translate `exclude_paths` to a `Gitattributes.pattern` and use the more generic mechanism. That way there's only one way to ignore paths.

lib/gitattributes.ml
```diff
@@ -0,0 +1,132 @@
...
+  | Prefix of string
+      (** Pattern like [dir/**] that matches everything under a directory, but
+          not the directory itself. *)
+  | Glob of Re.re  (** Compiled glob pattern. *)
+
+(** [glob_to_re pattern] is a compiled regex for glob [pattern]. Supports [*]
+    (any chars except /), [?] (single char except /), and [**] (any path
+    segments, but only when adjacent to /). *)
+let glob_to_re pattern =
+  let buf = Buffer.create (String.length pattern * 2) in
```
Leonidas-from-XIV, line 14: Why is the initial size `pattern * 2`?

lib/gitattributes.ml
```diff
@@ -0,0 +1,132 @@
...
+  Buffer.add_char buf '^';
+  let len = String.length pattern in
```
Leonidas-from-XIV, line 16: Nitpick but the length of the pattern is determined twice, so could be pulled up in the function an reused.

lib/gitattributes.ml
```diff
@@ -0,0 +1,132 @@
...
+  let rec loop i =
```
Leonidas-from-XIV, line 17: I wonder if it wouldn't make sense to use a `string list` and then `List.rev |> String.concat` instead of a mutable buffer here.

lib/gitattributes.ml
```diff
@@ -0,0 +1,132 @@
...
+          (* Escape regex metacharacters *)
+          Buffer.add_char buf '\\';
+          Buffer.add_char buf c;
+          loop (i + 1)
+      | _ ->
+          (* Literal character *)
+          Buffer.add_char buf c;
+          loop (i + 1)
+  in
+  loop 0;
```
Leonidas-from-XIV, line 58: Maybe it could be more maintainable to use `ocamllex` for this instead of hand-writing a tokenizer where the tokens are regex snippets.

lib/gitattributes.ml
```diff
@@ -0,0 +1,132 @@
...
+      (* Match everything under the directory, but not the directory itself *)
+      String.is_prefix ~affix:(prefix ^ "/") path_str
+  | Glob re ->
+      (* Match against full path or basename for patterns like *.log *)
+      Re.execp re path_str || Re.execp re basename
+
+let parse_export_ignore content =
+  (* Strip UTF-8 BOM if present at start of file *)
+  let content =
+    if String.is_prefix ~affix:"\xef\xbb\xbf" content then
```
Leonidas-from-XIV, line 105: I would suggest putting `"\xef\xbb\xbf"` into its own binding.

lib/gitattributes.ml
```diff
@@ -0,0 +1,132 @@
...
+      String.Sub.to_string (String.sub ~start:3 content)
```
Leonidas-from-XIV, line 106: And use `String.length utf8_bom` here.

lib/gitattributes.mli
```diff
@@ -0,0 +1,32 @@
...
+
+    Parses [.gitattributes] files and extracts patterns marked with the
+    [export-ignore] attribute. These patterns can be used to exclude files from
+    distribution archives. *)
+
+open Bos_setup
+
+(** {1 Patterns} *)
+
+type pattern
```
Leonidas-from-XIV, line 11:
```suggestion
type t
```

tests/lib/test_gitattributes.ml
```diff
@@ -0,0 +1,354 @@
...
+    integration without requiring git. *)
+let test_archive_export_ignore () =
+  let ( >>= ) = Result.bind in
+  let result =
+    Bos.OS.Dir.tmp "archive-test-%s" >>= fun dir ->
+    (* Create test file structure *)
+    let files =
+      [
+        ("CHANGES.md", "changes");
+        ("foo.opam", "opam");
```
Leonidas-from-XIV, line 24:
```suggestion
        ("foo.opam", {|opam-version: "2.0"|});
```

tests/lib/test_gitattributes.ml
```diff
@@ -0,0 +1,354 @@
...
+    in
+    let create_file (path, content) =
+      let fpath = Fpath.(dir // v path) in
+      let parent = Fpath.parent fpath in
+      Bos.OS.Dir.create ~path:true parent >>= fun _ ->
+      Bos.OS.File.write fpath content
+    in
+    List.fold_left
+      (fun acc file -> acc >>= fun () -> create_file file)
+      (Ok ()) files
```
Leonidas-from-XIV, line 44: I think `List.iter` followed by some kind of `Result.List.all_unit` (or the like) would be a bit easier to understand but it's of minor importance.

README.md
```diff
@@ -110,6 +110,31 @@ The full documentation of this command is available with
...
+```
+
+**Supported patterns:**
+- Exact filenames: `dune-workspace`
+- Directory patterns: `.github/**` (matches all files under `.github/`)
+- Glob patterns: `*.log`, `test_*`, `file?.txt`
+- Double star in path: `**/build`, `src/**/test.ml`
+- Path normalization: handles `./` and `../` in paths
+
+**Not supported:**
```
Leonidas-from-XIV, line 130: I think it would be good to state what "not supported" means in practice. Will it error? Will it ignore the pattern?

Leonidas-from-XIV (commented):
Very nice PR, the extensive test coverage is definitely a highlight and good to see as the glob to regex code is somewhat hairy.

I have some suggestions for improvements here.

lib/archive.ml
jberdine: Renamed. My thinking was that conceptually what is being done is "excluding" and so kept that base term, but fair point about double-negation.

lib/archive.ml
jberdine: My thinking is that the context here is reading a `.gitattributes` file, and so the expectation is to interpret it the way that `git archive` does (as much as possible). As a user, if some workflows used `git archive` and some `dune release`, and I had to adjust the `.gitattributes` file because they interpreted it differently, I would assume that `dune` was wrong since the name `.gitattributes` indicates that `git` "owns" the format.

lib/archive.mli
jberdine: That could make sense. My worry is that `exclude_paths` excludes based on basename, so at any depth, while `Gitattributes.pattern`s are path-anchored. Translating `exclude_paths` could have a semantic change if not done exactly right. So I think such a change ought to be in a separate PR, and I would want to have a much better idea for a good way to test it than I have at the moment.

lib/gitattributes.ml
jberdine: Dropped the `* 2`. It is a guess, based on the observation that each char of the pattern might expand into 2 regex chars, but the buffer grows fine so it is simpler to just skip the `* 2`.

lib/gitattributes.ml
jberdine: Fixed.

lib/gitattributes.ml
jberdine: That could be done. But I think that using a Buffer for building a string char-by-char is idiomatic, and the mutable state is encapsulated within a single function.

lib/gitattributes.ml
jberdine: Maybe, but I don't know. It would add a build dependency and generated file, and I don't expect that the syntax of globs is likely to change often. It could just as easily be done later, but I don't know if I will have time to work on that soon.

lib/gitattributes.ml
jberdine: Done.

lib/gitattributes.ml
jberdine: Done.

lib/gitattributes.mli
jberdine: Done.

tests/lib/test_gitattributes.ml
jberdine: Done.

tests/lib/test_gitattributes.ml
jberdine: Perhaps. I like the way the current formulation uses `>>=`/`Result.bind` to chain the results just like the rest of the code.

README.md
jberdine: Good point. I have pushed a commit that detects the unsupported pattern syntax, and updated the README to distinguish such cases (that are skipped with a warning) and other limitations (that are silent).

PR #517, opened 2026-03-24
Add a CLI flag for prerelease

Sudha247:
Fixes #516.

Adds a flag to mark a release as an aprerelease. Support for this already exists in the GitHub REST API. This patch exposes it to `dune-release` CLI.

PR #518, opened 2026-03-25
Explicitely pass flag to decompress bzip2 archives in test

Leonidas-from-XIV:
While looking at the CI results of #517 the (optional) [OpenBSD tests are failing](https://github.com/tarides/dune-release/pull/517/checks?check_run_id=68503371124). But it looks that the only thing preventing them from passing is one single cram test.

`tar` on OpenBSD can decompress bzip2 but it does not do that automatically by default, unlike GNU tar or bsdtar.

However all tar implementations support the `j` flag so might as well use it in the test - we know it's always going to be a `bzip2` compressed file, so the autodetection is not necessary.

This should fix tests on OpenBSD while keeping it working just fine on all other platforms too.

### DIFF
--- a/tests/bin/opam-file-locations/run.t
+++ b/tests/bin/opam-file-locations/run.t
@@ -100,5 +100,5 @@ And as well have a release tarball
 
 Which contains the `.opam` file in the right location:
 
-  $ tar tf _build/myproject-0.2.0.tbz | grep \\.opam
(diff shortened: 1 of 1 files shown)


shonfeder (approved):

Sudha247 (approved):
LGTM, for what it's worth!

